Issue #11, opened 2014-02-27
Highlight matched substring in results


DouweM:
Like in SublimeText:
![](http://cdn.tutsplus.com/net/uploads/legacy/1038_sublime/activate.slim.jpg)

This would also be useful in https://github.com/atom/command-palette.

mnquintana:
This should be implemented in [atom/atom-space-pen-views](https://github.com/atom/atom-space-pen-views) for `SelectListView` so the Command Palette can also take advantage of this.

izuzak:
> This should be implemented in atom/atom-space-pen-views for SelectListView so the Command Palette can also take advantage of this.

I'm not sure that should be the case (or that it's easily doable). Notice that the SelectListView is populated with HTML generated completely by subclases:

https://github.com/atom/atom-space-pen-views/blob/0e5fc262cc62d453bbd6dce9ecaec5caddc553f9/src/select-list-view.coffee#L85-L96

The generated view for an item doesn't need to contain the text you typed, for example, you could use it to show images and use the text you type to filter images somehow. Or it could contain some text that the user of the component doesn't want highlighted in this way. Maybe subclassing SelectListView might be a better fit?

mnquintana:
@izuzak Ah, you're right. Subclassing SelectListView would work better, but I'm not totally sure about the deprecation status of these Views so I'd be hesitant to do it.

mnquintana:
Added in fuzzy-finder v0.84.0 - should be in the next Atom release (0.199.0)!

PR #39, opened 2014-05-26
Highlight matches in search results

philschatz:
This adds highlighting matched text to the fuzzy search box and should address the issues described in https://github.com/atom/fuzzy-finder/issues/11.

Screenshot:

image

Animated:

fuzzy-overlap

**Note:** This adds `<strong>` around the matched characters but requires a change in CSS to show up white.

### DIFF
--- a/lib/fuzzy-finder-view.coffee
+++ b/lib/fuzzy-finder-view.coffee
@@ -79,10 +79,10 @@ class FuzzyFinderView extends SelectListView
         else
           typeClass = 'icon-file-text'
 
-        basename = path.basename(filePath)
-        baseOffset = projectRelativePath.length - basename.length
+        fileBasename = path.basename(filePath)
+        baseOffset = projectRelativePath.length - fileBasename.length
 
-        @div class: "primary-line file icon #{typeClass}", -> highlighter(basename, matches, baseOffset)
+        @div class: "primary-line file icon #{typeClass}", 'data-name': fileBasename, 'data-path': projectRelativePath, -> highlighter(fileBasename, matches, baseOffset)
         @div class: 'secondary-line path no-icon', -> highlighter(projectRelativePath, matches, 0)
 
   openPath: (filePath, lineNumber) ->
--- a/lib/fuzzy-finder-view.coffee
+++ b/lib/fuzzy-finder-view.coffee
@@ -45,6 +45,7 @@ class FuzzyFinderView extends SelectListView
         lastIndex = 0
         for matchIndex in matches
           matchIndex -= offsetIndex
+          continue if matchIndex < 0 # If marking up the basename, omit path matches
           unmatched = path.substring(lastIndex, matchIndex)
           matchedChar = path[matchIndex]
 
@@ -81,8 +82,8 @@ class FuzzyFinderView extends SelectListView
         basename = path.basename(filePath)
         baseOffset = projectRelativePath.length - basename.length
 
-        @div class: "primary-line file icon #{typeClass}", -> highlighter(basename, matches.base, baseOffset)
-        @div class: 'secondary-line path no-icon', -> highlighter(projectRelativePath, matches.path, 0)
+        @div class: "primary-line file icon #{typeClass}", -> highlighter(basename, matches, baseOffset)
+        @div class: 'secondary-line path no-icon', -> highlighter(projectRelativePath, matches, 0)
 
   openPath: (filePath, lineNumber) ->
     if filePath
--- a/lib/fuzzy-finder-view.coffee
+++ b/lib/fuzzy-finder-view.coffee
@@ -3,6 +3,7 @@ path = require 'path'
 {$$, SelectListView} = require 'atom-space-pen-views'
 {repositoryForPath} = require './helpers'
 fs = require 'fs-plus'
+{match} = require 'fuzzaldrin'
 
 module.exports =
 class FuzzyFinderView extends SelectListView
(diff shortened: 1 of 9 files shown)


brendanblackwood:
Would love to have this feature. Is there something blocking it right now?

dbpolito:
looking forward for this too..

izuzak:
I think this would be neat too. :+1: This PR no longer merges cleanly due to some minor conflicts, but assuming we can clean that up and add some tests -- @kevinsawicki how do you feel about adding this to fuzzy finder? (I noticed you looked at this PR previously.) 

We should also probably add the styles form https://github.com/atom/fuzzy-finder/pull/39#issuecomment-45975608 to the PR, but use something different than `color: white` since that won't look good in all themes?

mnquintana:
This would be really awesome! I filed a related issue in atom/atom-space-pen-views#10. I think this should live in atom/atom-space-pen-views instead of in fuzzy-finder though, that way the Command Palette can also make use of this enhancement (and anything else that extends `SelectListView`.

philschatz:
Forgot to actually push these changes up last week :blush: . Hope this helps! And, if you guys do another Musical Desks event I'd love to swing by and get some more stickers!

lib/fuzzy-finder-view.coffee
```diff
@@ -33,7 +34,29 @@ class FuzzyFinderView extends SelectListView
...
+      highlighter = (path, matches, offsetIndex) =>
+        lastIndex = 0
+        for matchIndex in matches
+          matchIndex -= offsetIndex
+          continue if matchIndex < 0 # If marking up the basename, omit path matches
+          unmatched = path.substring(lastIndex, matchIndex)
+          matchedChar = path[matchIndex]
+
+          @text unmatched if unmatched
+          @strong matchedChar, class: 'matched-char'
```
mnquintana: I think this should be a `span` instead – this would allow for finer grained control over styling. (Bold / strong text isn't necessarily very visible in every theme.) Could you also add a style for this class in `styles/`?

mnquintana: Also, what do you think about joining adjacent matched chars into one `'.matched-chars'` element? I think this might be better for semantics / styling.

mnquintana: Or, actually, it could still be a strong tag, just with custom styling for that class.

philschatz: Yep, I can switch it to `<span>` and try to join adjacent matched chars. Where is `styles/` and is there a better name than `.matched-char`?

mnquintana: Ah sorry - you'll need to create a `styles/` folder in the root of the package. Hmmm... maybe something like `.match` or `.matching`?

philschatz:
I believe most of @mnquintana's issues are resolved. Here is a screenshot with brackets in CSS to show that sets of characters are matched:
image

- [x] use `<span>` instead of `<strong>`
- [x] group matched characters to be more semantic
- [ ] add styles (really just any old file in `/styles`?)

And the overspecified CSS:

```less
div.select-list ol.list-group li.two-lines span.matching {
  color: white;
  &::before { content: '['; }
  &::after  { content: ']'; }
}
```

mnquintana:
@philschatz This is looking great! I'll check out the code later today, but just a few points:
- Can you add specs for this? We wouldn't want to break this functionality accidentally in the future.
- For the styles, you can just add a `fuzzy-finder.less` file (just a convention, it could really be called anything, but it should be a Less file) in `styles/`.
- I'm not so sure about the brackets for the final styling - it's a bit obtrusive. Maybe just a simple underline for the default style would suffice.
```less
div.select-list ol.list-group li.two-lines span.matching {
  color: white;
}
```
The only problem with specifying a color like this is that it means it won't be themeable by UI themes - it'll be `white` no matter what UI theme is applied. To make the color adjust with the UI theme, you can `@import "ui-variables"` in the `.less` file you'll create, and you can make use of [one of these variables](https://atom.io/docs/latest/hacking-atom-creating-a-theme#variables) that UI themes will provide, to make the color more dynamic.

philschatz:
Yep, I can add specs and a LESS file later. The CSS I provided was only meant to illustrate that matched characters were grouped; brackets would be quite ugly indeed!

philschatz:
image

Added specs and LESS file.

I was unsure of the scope of tests and CSS selector specificity but I'm happy to make changes!

spec/fuzzy-finder-spec.coffee
```diff
@@ -624,6 +624,65 @@ describe 'FuzzyFinder', ->
           expect(atom.workspace.getActiveTextEditor()).toBe editor1
           expect(editor1.getCursorBufferPosition()).toEqual [3, 4]
 
+  describe "highlights matches", ->
```
50Wliu: How about "match highlighting"?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fuzzy-finder",
-  "version": "0.70.0",
+  "version": "0.72.0",
```
mnquintana: Not sure what happened here, but it looks like this is an old version of package.json - can you try merging / rebasing master again? The version should be 0.81.0

philschatz:
Merged from master and updated the test description

philschatz:
Howdy! Is this waiting on something from me?

The current look (using `@text-color-highlight`):
image

mnquintana:
@philschatz Thanks for the ping! This looks good to me - thanks for adding this :ship:

mnquintana:
(Also, sorry this took so long to merge :green_heart:)

philschatz:
Given the [1-year anniversary](https://github.com/blog/1997-atom-1-year-open-source-anniversary) last week I wonder if this wins as the:

## :tada: :balloon: :trophy: Longest Pull Request :balloon: :tada: 

I'd love to do some more open co-coding over a :beer: the next time you guys host an event (SF or PDX)!

Also, good luck @kevinsawicki on GSoC; I've been a mentor for several years and it's a ton of fun!

kevinsawicki:
> Given the 1-year anniversary last week I wonder if this wins

Yeah, I apologize for not merging this sooner and losing track of it over the last year :chart_with_downwards_trend: 

Fortunately we've added some great collaborators in the past year like @mnquintana and I'm optimistic that awesome contributions like this won't be neglected in the future :chart_with_upwards_trend: 

> I'd love to do some more open co-coding over a :beer: the next time you guys host an event (SF or PDX)!

Yeah, I'm hoping to go to PDX in July for OSCON so meeting up then would be cool.

maxbrunsfeld:
@philschatz Thanks so much for adding this. Love it.

PR #73, opened 2015-02-09
Add feature to search through all open panes instead of just active pane

### LINKED ISSUE #34: Choose already opened file even from another pane
(This could be a fuzzy-finder option. Original issue is https://github.com/atom/atom/issues/1303)

## Behavior expected

Opening a file from cmd-b or cmd-p should go directly to the file if it's already opened, regardless of pane.

## Behavior observed

The file is selected if you're in the same pane, but opened new if not.

## Thoughts

This could be behavior that can be had both ways. For example, if you open a file with enter it will go to the already opened file but if you open with shift-enter it opens a new copy in the current pane.

/via @caseycrites

tmunro:
Fixes #34: Add feature to search through all open panes instead of just the active one, a preference to control this, and the ability to invert the current preference with Shift-Enter.


izuzak:
Thanks for opening this, @tmunro! It seems that this PR no longer merges cleanly -- would you be willing to update it so that it is mergeable? 

Also, could you add some tests for the new behavior? That would prevent future contributors from breaking the behavior.

tmunro:
I can definitely do that - I'll try to set aside some time this weekend to clean it up.

PR #78, opened 2015-03-03
Remove ctrl-t keybinding for showing fuzzy finder on Linux and Windows

izuzak:
See https://github.com/atom/atom/pull/3050#issuecomment-49906731 for motivation. On OSX, `cmd+t` and `cmd+p` are still valid, and that matches the behavior in SublimeText (both keybinding are valid there as well on OSX, and for Linux/Win only `ctrl-p` is valid).

cc @kevinsawicki

### DIFF
--- a/keymaps/fuzzy-finder.cson
+++ b/keymaps/fuzzy-finder.cson
@@ -5,13 +5,11 @@
   'cmd-B': 'fuzzy-finder:toggle-git-status-finder'
 
 '.platform-win32':
-  'ctrl-t': 'fuzzy-finder:toggle-file-finder'
   'ctrl-p': 'fuzzy-finder:toggle-file-finder'
   'ctrl-b': 'fuzzy-finder:toggle-buffer-finder'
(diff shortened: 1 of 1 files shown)


50Wliu:
Hmm.  I disagree with this because my first instinct was to use Ctrl+T because I equated it with "New tab" :P.
Though I guess I'll get used to it.

izuzak:
@50Wliu Yeah, I get that. I'm pretty used to `cmd-t` on OSX as well so I'd probably be :angry: if that was changed. We could also go the other direction and leave these as they are and just use `ctrl-alt-t` for transpose in https://github.com/atom/atom/pull/3050, as suggested in https://github.com/atom/atom/pull/3050#issuecomment-49904502 (which wouldn't match Sublime, but we wouldn't be making our Windows/Linux users who are used to `ctrl-t` angry :goberserk:). Gonna leave this open for a bit longer to get feedback from more folks. Thanks!

izuzak:
Gonna close this since it's been almost 4 years and I probably won't be the one pushing this over the 🏁 finish line 🏁

PR #90, opened 2015-03-28
Add option to preserve last search query

zigomir:
When working, I love having minimal number of opened tabs. Thus, I close and open files all the time. This is kinda ingrained habit for me. I'm really used to flow of JetBrains products, because their fuzzy search remembers a query string of your previous search.

Lack of this option was always a deal breaker for me. But since Atom is so hackable, it was just few lines of code to make it work. What's your opinion on it?

fuzzy

Here's a gif where I show pain of typing same query string two times. And then, after enabling this option where you can see the preserved query string so I can open file instantly. From this example it might seem I'm not aware of undo close tab shortcut. This is not the case. 

Better way to describe problem that this feature solves is that on our projects we have lots of files with similar names (models / views). When you open a model file you later on want to open view file too and you don't want to type same file name again.

### DIFF
--- a/lib/fuzzy-finder-view.coffee
+++ b/lib/fuzzy-finder-view.coffee
@@ -28,6 +28,16 @@ class FuzzyFinderView extends SelectListView
   getFilterKey: ->
     'projectRelativePath'
 
+  cancel: ->
+    if atom.config.get('fuzzy-finder.preserveLastSearch')
+      lastSearch = @getFilterQuery()
+      super
+
+      @filterEditorView.setText(lastSearch)
+      @filterEditorView.getModel().selectAll()
+    else
+      super
+
   destroy: ->
     @cancel()
     @panel?.destroy()
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -6,6 +6,9 @@ module.exports =
     traverseIntoSymlinkDirectories:
       type: 'boolean'
       default: false
+    preserveLastSearch:
+      type: 'boolean'
+      default: false
 
   activate: (state) ->
     atom.commands.add 'atom-workspace',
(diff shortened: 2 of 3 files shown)


zigomir:
Any feedback? Should this be an additional package?

kevinsawicki:
> Any feedback? Should this be an additional package?

I like this and think it makes sense to include this as an option in this package like you have it now.

Would you mind adding a spec for this new option?

zigomir:
@kevinsawicki thanks for feedback. I added specs.

kevinsawicki:
Sorry, looks like this no longer cleanly merges, would you mind rebasing/merging master into this branch?

zigomir:
@kevinsawicki done

zigomir:
@kevinsawicki did you change your mind about this feature?

kevinsawicki:
> did you change your mind about this feature?

Nope, just lost track of this pull request, will re-review and merge shortly, thanks for the bump :+1:

lib/fuzzy-finder-view.coffee
```diff
@@ -30,6 +30,16 @@ class FuzzyFinderView extends SelectListView
   getFilterKey: ->
     'projectRelativePath'
 
+  cancel: ->
+    if atom.config.get('fuzzy-finder.preserveLastSearch')
+      lastSearch = @getFilterQuery()
```
kevinsawicki: calling `@getFilterQuery()` returns the filter text without a trailing line number such so `pack:8` would become `pack`.

Did you want to preserve this info across opens? A trailing `:NUMBER` allows you to open a file to a specific line.

kevinsawicki: I think we should ship this as-is to start, can always tweak it later.

kevinsawicki:
Thanks for this :shipit:

PR #92, opened 2015-04-09
Deduplicate fuzzy finder keybindings

thedaniel:
This will require a docs update, but opening for discussion as this came up in chat with @bolinfest.

Good keybindings are precious and rare, and we don't have to hew perfectly to sublime for new users. I understand that users can rebind commands as necessary, but as most users will likely stick to the defaults, I think we should pick one binding for the fuzzy finder and stick with it. I personally remapped cmd-p to the command palette literally the first day I started using Atom so I prefer to keep cmd-t but I know that's totally subjective and could be argued into doing this the other way.

cc: @bolinfest , @atom/core

### DIFF
--- a/keymaps/fuzzy-finder.cson
+++ b/keymaps/fuzzy-finder.cson
@@ -1,17 +1,14 @@
 '.platform-darwin':
   'cmd-t': 'fuzzy-finder:toggle-file-finder'
-  'cmd-p': 'fuzzy-finder:toggle-file-finder'
   'cmd-b': 'fuzzy-finder:toggle-buffer-finder'
   'cmd-B': 'fuzzy-finder:toggle-git-status-finder'
 
 '.platform-win32':
   'ctrl-t': 'fuzzy-finder:toggle-file-finder'
-  'ctrl-p': 'fuzzy-finder:toggle-file-finder'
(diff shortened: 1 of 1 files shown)


mnquintana:
I am extremely biased here, but I came to Atom directly from Sublime, so I favor the <kbd>cmd-p</kbd> shortcut (although I came to Sublime from TextMate, so I was used to <kbd>cmd-t</kbd> before... *shrugs*).

lee-dohm:
One other data point is that the GitHub website uses the `t` shortcut to go to the file picker. Consistency with other GitHub products should count for something too.

bolinfest:
Ping? Let's move this forward one way or the other to free up a keybinding!

nathansobo:
I'm honestly reluctant to remove either binding, but `cmd-t` should probably be the one to die if we really need to do it.

mehcode:
:+1: :100: 

The `ctrl-` and `cmd-` keybindings are a precious resource. Having two for the same package is :panda_face:.

However I agree that we'd need to remove `cmd-t` / `ctrl-t`
 - We have `ctrl-shift-p` for the command panel
 - We could have `ctrl-alt-p` for a project switcher (the spec runner doesn't need a keyboard shortcut as the average user doesn't care)
 - It mimics sublime in a very obvious way

PR #95, opened 2015-04-16
:racehorse: Optimize realpath checks

maxbrunsfeld:
* Use a cache
* Use async realpath

### DIFF
--- a/lib/load-paths-handler.coffee
+++ b/lib/load-paths-handler.coffee
@@ -80,7 +80,7 @@ class PathLoader
       if err
         done(false)
       else
-        done(realPath?.search(@rootPath) is 0)
+        done(realPath.search(@rootPath) is 0)
 
 module.exports = (rootPaths, followSymlinks, ignoreVcsIgnores, ignores=[]) ->
   ignoredNames = []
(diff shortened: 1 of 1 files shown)


lib/load-paths-handler.coffee
```diff
@@ -74,13 +75,12 @@ class PathLoader
...
-      realPath = fs.realpathSync(pathToLoad)
-    catch err
-      ; # ignore a broken symlink
-    realPath?.search(@rootPath) is 0
+  isInternalSymlink: (pathToLoad, done) ->
+    fs.realpath pathToLoad, @realPathCache, (err, realPath) =>
+      if err
+        done(false)
+      else
+        done(realPath?.search(@rootPath) is 0)
```
kevinsawicki: I don't think `realPath` can be `null` at this point so the `?` is probably no longer needed.

maxbrunsfeld:
This optimization doesn't make much of a difference on my machine. I never noticed any slowness to begin with, and even in a directory with lots of symlinks (my `~/.atom`), indexing was already fast without this. But it might make a difference on a machine with an HDD.

PR #97, opened 2015-04-26
Add feature to search through all open panes instead of just active pane

### LINKED ISSUE #34: Choose already opened file even from another pane

tmunro:
Fixes #34: Add feature to search through all open panes instead of just the active one, a preference to control this, and the ability to invert the current preference with Shift-Enter.

@izuzak Sorry about creating a second PR, I got confused about how to rebase and screwed up the other branch.

### DIFF
--- a/lib/fuzzy-finder-view.coffee
+++ b/lib/fuzzy-finder-view.coffee
@@ -24,6 +24,8 @@ class FuzzyFinderView extends SelectListView
         @splitOpenPath (pane, item) -> pane.splitDown(items: [item])
       'pane:split-up': =>
         @splitOpenPath (pane, item) -> pane.splitUp(items: [item])
+      'fuzzy-finder:alternate-confirm': =>
+        @confirmAlternateSelection()
 
   getFilterKey: ->
     'projectRelativePath'
@@ -61,9 +63,10 @@ class FuzzyFinderView extends SelectListView
         @div fileBasename, class: "primary-line file icon #{typeClass}", 'data-name': fileBasename, 'data-path': projectRelativePath
         @div projectRelativePath, class: 'secondary-line path no-icon'
 
-  openPath: (filePath, lineNumber) ->
+  openPath: (filePath, lineNumber, searchAllPanes) ->
     if filePath
-      atom.workspace.open(filePath).done => @moveToLine(lineNumber)
+      if not atom.workspace.getActivePane().activateItemForURI(filePath)
+          atom.workspace.open(filePath, searchAllPanes: searchAllPanes ).done => @moveToLine(lineNumber)
 
   moveToLine: (lineNumber=-1) ->
     return unless lineNumber >= 0
@@ -100,9 +103,13 @@ class FuzzyFinderView extends SelectListView
 
   confirmSelection: ->
     item = @getSelectedItem()
-    @confirmed(item)
+    @confirmed(item, atom.config.get 'fuzzy-finder.searchAllPanes')
 
-  confirmed: ({filePath}={}) ->
+  confirmAlternateSelection: ->
+    item = @getSelectedItem()
+    @confirmed(item, !atom.config.get 'fuzzy-finder.searchAllPanes')
+
+  confirmed: ({filePath}={}, searchAllPanes) ->
     if atom.workspace.getActiveTextEditor() and @isQueryALineJump()
       lineNumber = @getLineNumber()
       @cancel()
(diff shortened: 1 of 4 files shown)


lib/fuzzy-finder-view.coffee
```diff
@@ -61,9 +63,12 @@ class FuzzyFinderView extends SelectListView
...
         @div projectRelativePath, class: 'secondary-line path no-icon'
 
-  openPath: (filePath, lineNumber) ->
+  openPath: (filePath, lineNumber, searchAllPanes) ->
     if filePath
-      atom.workspace.open(filePath).done => @moveToLine(lineNumber)
+      if atom.workspace.getActivePane().activateItemForURI(filePath)
+        @moveToLine(lineNumber)
+      else
+        atom.workspace.open(filePath, searchAllPanes: searchAllPanes ).done => @moveToLine(lineNumber)
```
izuzak: Why is this if/else block needed and you can't use this for both cases:

```coffee
atom.workspace.open(filePath, searchAllPanes: searchAllPanes ).done => @moveToLine(lineNumber)
```

lib/fuzzy-finder-view.coffee
```diff
@@ -24,6 +24,8 @@ class FuzzyFinderView extends SelectListView
         @splitOpenPath (pane, item) -> pane.splitDown(items: [item])
       'pane:split-up': =>
         @splitOpenPath (pane, item) -> pane.splitUp(items: [item])
+      'fuzzy-finder:alternate-confirm': =>
```
mnquintana: This name is a bit vague - I wasn't quite sure what it did just from looking at the name. Maybe something more like `invert-confirm`? (That's probably not a great name either, but I think it should change.)

spec/fuzzy-finder-spec.coffee
```diff
@@ -404,6 +404,9 @@ describe 'FuzzyFinder', ->
           dispatchCommand('toggle-buffer-finder')
 
       describe "when the active pane has an item for the selected path", ->
+        beforeEach ->
+          atom.config.set("fuzzy-finder.searchAllPanes", false)
```
izuzak: Since `false` is the default value for this config option, is this call to `atom.config.set` needed here? I'm wondering if you had a specific reason for adding that in this test.

izuzak:
@tmunro Thanks! I think this looks good, just had a few comments (thanks @mnquintana! :metal:). 

@nathansobo -- you commented on #34 which this PR will close. Do you have some :hourglass: to take a look at this as well?

spec/fuzzy-finder-spec.coffee
tmunro: No good reason - just unfamiliarity with the way these tests handle state. Removed.

lib/fuzzy-finder-view.coffee
tmunro: I swear there was a reason, but it works and tests just fine simplified. Changed.

lib/fuzzy-finder-view.coffee
tmunro: Changed to `invert-confirm`, but I agree that it could be made clearer still. How important is brevity? Could be something like `invert-pane-search-preference-confirm`?

kevinsawicki:
Thanks for this, I'm going to merge it and then make a few minor changes on master and will link the commit here so you can see what I tweaked.

kevinsawicki:
Here were the tweaks I made, really minor stuff: https://github.com/atom/fuzzy-finder/commit/b44656b239ffb0566465c8e8efffc2f51855d229

Typically in packages and core we try to pass around options objects instead of boolean parameters so they are easier to understand when reading lines calling those methods such as:

```coffee
@openPath(filePath, false)
```

versus:

```coffee
@openPath(filePath, searchAllPanes: false)
```

tmunro:
Good to know! Happy to help in my own small way, thanks for merging and making those changes.

kevinsawicki:
I also tweaked the `fuzzy-finder:invert-confirm` keybinding to only be enabled on the mini editor inside the fuzzy finder view. Previously it was registered globally.

https://github.com/atom/fuzzy-finder/commit/9b856d2155c7501ebb81e3adab00f0d6416407bf

PR #103, opened 2015-05-19
Fix and optimize load-paths-handler.coffee

pgkos:
My initial attempt at fixing atom/atom#6437. It should also make indexing much faster.


maxbrunsfeld:
It looks like this change causes a lot of tests to fail. Are you able to run the tests locally and debug the failures?

pgkos:
Yeah, I have made a mistake in PathLoader.isPathIgnored (parameter name conflict with `path`, fixed in the latest commit), but the tests still fail.

The same tests unfortunately also fail when I run them on the original version of fuzzy-finder shipped with Atom (at least on my PC, Ubuntu 64-bit).

pgkos:
I have (hopefully) fixed all the issues

PR #104, opened 2015-05-27
Add coffeelint support

mnquintana:
Refs https://github.com/atom/atom/issues/5982


mnquintana:
atom.io's having trouble right now, but specs pass locally

PR #106, opened 2015-06-01
Remove 'cmd-t' / 'ctrl-t' for 'toggle-file-finder'

mehcode:
Replaces #92 and #78 
See https://github.com/atom/fuzzy-finder/pull/92#issuecomment-107597085

----

The `ctrl-` and `cmd-` keybindings are a precious resource. Having two for the same package is :panda_face:.

However I agree that we'd need to remove `cmd-t` / `ctrl-t`
 - We have `ctrl-shift-p` for the command panel
 - We could have `ctrl-alt-p` for a project switcher (the spec runner doesn't need a keyboard shortcut as the average user doesn't care)
 - It mimics sublime in a very obvious way

### DIFF
--- a/keymaps/fuzzy-finder.cson
+++ b/keymaps/fuzzy-finder.cson
@@ -1,17 +1,14 @@
 '.platform-darwin':
-  'cmd-t': 'fuzzy-finder:toggle-file-finder'
   'cmd-p': 'fuzzy-finder:toggle-file-finder'
   'cmd-b': 'fuzzy-finder:toggle-buffer-finder'
   'cmd-B': 'fuzzy-finder:toggle-git-status-finder'
 
 '.platform-win32':
-  'ctrl-t': 'fuzzy-finder:toggle-file-finder'
   'ctrl-p': 'fuzzy-finder:toggle-file-finder'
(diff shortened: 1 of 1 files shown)


lee-dohm:
Before this is merged, we should probably scour the documentation to ensure that there are no references to <kbd>Cmd/Ctrl+T</kbd>.

mehcode:
@lee-dohm The welcome screen and the background tips come to mind as well.

mnquintana:
Also just a note for after this is merged: we should probably add a note in the release notes for the version of Atom that incorporates this change about how to re-add this keybinding for those who miss it (and you know there will be people who'll miss it)

50Wliu:
Still :-1: on removing the -T keybinding for the reasons listed in #92.  Ctrl+T is newtab in every major web browser, t is the fuzzy finder in Github...

mehcode:
The big reason that sways me in favor of `cmd-p` over `cmd-t` is that all the similar bindings you mention are `new tab/document/file`, `find file`. `cmd-p` is far more than just `find file`, it's not quite to sublime's level of "goto anything" but its close.

---

With that said, I really don't care. I just want to get rid of one of them. I don't mind updating the PRs if the decision is in the reverse.

PR #117, opened 2015-07-09
Complete the bindings, try to describe shift-enter

bronson:
Search All Panes is a difficult concept to explain succinctly but I tried...  Open to any suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,10 +5,10 @@ Quickly find and open files using `cmd-t`.
   * `cmd-t` or `cmd-p` to open the file finder
   * `cmd-b` to open the list of open buffers
   * `cmd-shift-b` to open the list of Git modified and untracked files
-  * `enter` opens the selected file without leaving the current pane (depending on the Search All Panes setting)
-  * `shift-enter` opens the file, switching to another pane if it's already open there. (depending on the Search All Panes setting)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,8 +2,12 @@
 
 Quickly find and open files using `cmd-t`.
 
-  * `cmd-b` to open the buffer list
+  * `cmd-t` or `cmd-p` to open the file finder
+  * `cmd-b` to open the list of open buffers
   * `cmd-shift-b` to open the list of Git modified and untracked files
+  * `enter` or `shift-enter` selects a file
+
+If you select a file, and it isn't open in the current pane, it will be opened in a new tab.  If it is already open in a different pane, you can hit `shift-enter` to switch to that pane instead of creating a new tab.  The "Search All Panes" setting reverses this, so enter opens the file in any pane and shift-enter creates a new tab in the current pane.
```
lee-dohm: I'd specify which key does what when "Search All Panes" is in its default state and state what the default state is, then state that switching it to the non-default state reverses it. Just from reading this, I'm not sure which way to set the setting to make it one or the other.

bronson:
@lee-dohm agreed!  Better?  And definitely tell me if you have formatting advice.

lee-dohm:
Thanks!

PR #128, opened 2015-08-07
Remove leading './'

amiuhle:
Rspec reports failed examples as

```
./path/to/file.rb
```

When double clicking on the path in order to copy it, `./` gets copied too, which is why I would like to have it filtered out in the fuzzy finder.
#### TODOs:
- [ ] Currently, '`./file` will find `file.rb`, even without the path. Should this only find complete paths?
- [ ] I added a spec that didn't pass before, but passes now, so I'm assuming that's okay. However, in other specs there are more expectations, so I'm not sure if I should check if the correct file is opened etc.

### DIFF
--- a/lib/fuzzy-finder-view.coffee
+++ b/lib/fuzzy-finder-view.coffee
@@ -173,6 +173,8 @@ class FuzzyFinderView extends SelectListView
     query = super
     colon = query.indexOf(':')
     query = query[0...colon] if colon isnt -1
+    # Remove leading './'
+    query = query.replace(/^\.\//, '')
     # Normalize to backslashes on Windows
     query = query.replace(/\//g, '\\') if process.platform is 'win32'
     query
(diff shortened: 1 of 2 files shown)


igas:
@as-cii is there reason why we can't do it? I saw few issues and found this MR, but no responses in them if there are something preventing us from having this feature.

as-cii:
I believe this change could be reasonable and I cannot think of any possible drawbacks. 

Before reviewing this, though, I think it would be good to rebase it on top of master (or even create a new branch, considering this repository has changed a lot since this pull request was opened).

@atom/maintainers: what do you think? 💭

lee-dohm:
What problem is this trying to solve? Are some files not being found until the `./` is removed?

igas:
@lee-dohm yes, it's valid unix path, when you get this path from 3rd party app/output and try to put in fuzzy finder you need to remove `./`

amiuhle:
@lee-dohm Yes, some tools report paths with a leading `./`. When copy-and-pasting the path into the fuzzy finder, those files will not be found without removing it first.

Do you want me to rebase this on `master`?

dedman:
This also annoys me, if I fix the conflicts would the PR be accepted?

lee-dohm:
Thanks for the pings on this.

In our weekly triage meeting yesterday, the maintainers decided that this violates the Principle of Least Surprise. While this may be a very obvious and useful change for you, there are a couple million other Atom users for whom this may not be obvious or benign. Because of this, we have decided to not accept this change.

Thank you all very much for your feedback and the work invested in this pull request.

PR #132, opened 2015-08-22
Option to show closed buffers too

viddo:
remember

This introduces an additional (opt-in) config setting to show be able to see recently-used-but-closed-buffers in the buffer view. The config setting defines how many closed buffers to remember (0 by default, i.e. the buffer view works just like now).

This is more or less the same functionality as I implemented in the package [recent-files-fuzzy-finder](https://github.com/viddo/recent-files-fuzzy-finder/blob/master/lib/recent-files-view.coffee), but considering I had to do some [workarounds/hacks](https://github.com/viddo/recent-files-fuzzy-finder/blob/master/lib/recent-files-view.coffee#L2-L6) to extend the buffer view I think it makes more sense to just add this as an additional (opt-in) behaviour in this package.

For some motivation to this feature see [recent-files-fuzzy-finder/readme#why](https://github.com/viddo/recent-files-fuzzy-finder#why)

### DIFF
--- a/lib/buffer-view.coffee
+++ b/lib/buffer-view.coffee
@@ -3,6 +3,10 @@ FuzzyFinderView = require './fuzzy-finder-view'
 
 module.exports =
 class BufferView extends FuzzyFinderView
+
+  initialize: (@closedBuffers) ->
+    super
+
   toggle: ->
     if @panel?.isVisible()
       @cancel()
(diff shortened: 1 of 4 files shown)


tnrich:
Any reason this one wasn't merged? It seems like a very nice feature to be able to have!

PR #147, opened 2015-10-27
Feature Save Search Across fuzzy finder menus

laginha87:
Keeps the search query when switching between File Finder, Buffer Finder and Git Status Finder.

I find it helpful when I'm using the 'File Finder', I remember that the file I'm looking for has been changed so I switch to 'Git Status Finder' to narrow the search but I have to start the query all over.

Added as a configuration option.

obligatory-gif-1
obligatory-gif-2

### DIFF
--- a/lib/fuzzy-finder-view.coffee
+++ b/lib/fuzzy-finder-view.coffee
@@ -238,6 +238,12 @@ class FuzzyFinderView extends SelectListView
   setItems: (filePaths) ->
     super(@projectRelativePathsForFilePaths(filePaths))
 
+  setText: (searchText)->
+    @filterEditorView.setText(searchText)
+
+  getText: ()->
+    @filterEditorView.getText()
+
   projectRelativePathsForFilePaths: (filePaths) ->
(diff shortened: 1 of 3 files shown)


thedaniel:
This seems like a pretty niche feature to me, especially since it changes the fast "hit hotkey and start typing" nature of the fuzzy-finder. I'm not sure it's worth adding a configuration option for. Perhaps just listing the most recent entry at the top of the result list when the query is empty would serve the same feature?

@atom/feedback any thoughts?

chick-chains:
@thedaniel To be clear this only triggers when you switch between different fuzzy finder windows. When you open the fuzzy finder window without anything open the text input is empty.
Also It seemed to me like a feature very similar to the preserve search option, that's why I added the configutration option.

benogle:
I'm also not super keen on this. What about all the other select list based UIs? If this were enabled, I would expect the term to be preserved between fuzzy-finder, command palette, bookmark searching, symbols view, etc.

chick-chains:
The list based UIs in the fuzzy finder package share a common context, they are all acting on files.

PR #149, opened 2015-10-31
Add down/up movement in fuzzy-finder list using tab/shift-tab

kandros:
I recently discovered this behaviour in the chrome address bar and find it incredibly usefull compared to using arrow of additional character


lee-dohm:
I'm not sure this should be the default but I'm not sure it shouldn't either. I don't *think* it will conflict with any of the various fuzzy-finder behaviors ...

/cc @atom/feedback

PR #151, opened 2015-11-06
Current opened files on top of list

laskos:
We have already last open file on top of files list. But what when i want switch to file which open before last one? This is the most missing feature from Sublime Text. After merge this PR we are able to switch between opened files just by `cmd+t`.


lee-dohm:
If you want a file that's currently open, you could use <kbd>Cmd+B</kbd> instead.

laskos:
@lee-dohm yes it works, but why we can't marge these two lists? I don't think, that someone scroll first view of `cmd-t` for looking something other than last opened files, like in ST.

PR #154, opened 2015-11-20
Add vim-mode keybinding for showing fuzzy finder

maciejpi:
Please review: Adding vim mode keybindings for: file finder, buffer finder, and git-status finder.


lee-dohm:
I believe this should go in the [vim-mode package](https://github.com/atom/vim-mode), that way all vim-mode keybindings can be maintained in one location.

/cc @maxbrunsfeld

maxbrunsfeld:
These aren't standard vim key bindings. I think you should just put this in your personal `keymap.cson`, @maciejpi.

PR #159, opened 2015-12-07
Use fuzz-aldrin-plus by default

nathansobo:
Heads up @benogle just so you know I'm shepherding this through.

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -14,7 +14,7 @@ module.exports =
       description: 'Remember the typed query when closing the fuzzy finder and use that as the starting query next time the fuzzy finder is opened.'
     useAlternateScoring:
       type: 'boolean'
-      default: false
+      default: true
       description: 'Use an alternative scoring approach which prefers run of consecutive characters, acronyms and start of words. (Experimental)'
 
   activate: (state) ->


benogle:
:+1:

jeancroy:
Is this still scheduled for 1.4 stable ?
I think it [missed the boat](https://github.com/atom/atom/blob/beta/package.json#L91) for 1.4 beta.

Also I'd like to inquiry about how are dependency updated ? Package.json state `fuzzaldrin-plus: ^0.1.0` which I guess means every minor version. But it looks you have a more fine grained control over this.

Finlay I'd like to know what need to happens on my side to issue update at the alpha /  in time for beta stage of test. 

Thanks you.

nathansobo:
Shoot. Lost track of things and forgot to bump the version in atom/atom. Sorry. We could hotfix this on 1.4.0-beta.

nathansobo:
Okay, I have this on the current beta branch. We have a couple other things to bundle into the next hotfix but this should be out soon.

benogle:
> We could hotfix this on 1.4.0-beta.

Also, can the same upgrade be applied to autocomplete-plus? Looks like the [default is still false](https://github.com/atom/autocomplete-plus/blob/96d377244d24c3e86309f35470db0232d0dbc4ff/lib/main.coffee#L121). This is a weird forum to ask, but in the same vein.

lib/main.coffee
```diff
@@ -14,7 +14,7 @@ module.exports =
       description: 'Remember the typed query when closing the fuzzy finder and use that as the starting query next time the fuzzy finder is opened.'
     useAlternateScoring:
       type: 'boolean'
-      default: false
+      default: true
       description: 'Use an alternative scoring approach which prefers run of consecutive characters, acronyms and start of words. (Experimental)'
```
mrfoto: Well if it's now `default` it shouldn't be indicated by `Experimental`, right?

benogle: word

jeancroy: Then you may be interested to remove the word `experimental` next to autocomplete-plus Symbol provider. I think it's the default since big 1.0

https://github.com/atom/autocomplete-plus/blob/931841dfc3a3318cbb141f5c7abc5eee50aaeb28/lib/main.coffee#L101-L105

Issue #70, opened 2015-02-02
Select and open multiple files at once


doggan:
In some other fuzzy finders that I've used in the past, you can use `shift + down` to select multiple files, and then press enter to open them all at once. Here is a quick mockup of what it would look like:

![mockup](https://cloud.githubusercontent.com/assets/517857/5996258/7fece128-aae9-11e4-9e6e-ab6bc32a65d5.gif)

This is particularly useful for opening related files that only differ by extension (`MyClass.h/.cpp`, `Options.html/.js/.css`, etc).

Thanks! :sparkles:

jrjohnson:
:+1:

hannupekka:
:+1: for this. With ST3 it is also possible to open a file with right arrow, while keeping search results still open.

admpsktt:
:+1:

admpsktt:
As @hannupekka mentioned, having the ability to open multiple files but not necessarily in the same order as they appear in the search results would be great.

admpsktt:
If I'm not mistaken, I think [Helm](https://github.com/emacs-helm/helm) has a feature where you can mark files to be opened and do so simultaneously once you've made your selection(s).

freemanoid:
I think it would be super useful to implement either opening a file from suggestions without closing search results or selecting several files to open (e.g. mark a file to open with some shortcut) or even better to have both possibilities.

svewag:
:+1:

slobo:
:+1:

BenjaminConant:
:+1:

JFRC:
:+1:
It's a useful feature.

intellix:
Came here to request the "right arrow" functionality from Sublime 3. Shift+down would achieve the same thing in some cases but when you have the ability to press right arrow, you can at least skip items you don't want to open.

if I've got 10 projects and I want to edit the bower.json within them all, I have to cmd+p and search 10 times. The most annoying part of this is that the search results are randomised each time so I have to open them in a particular project order to make sure I've opened all of the files I need.

Would be good to:

 - cmd + p > bower.json
 - go down and the list and press right-arrow x 10

indigo0086:
To illustrate how the current implementation is quite wasteful.
```
# of key presses  to open n files in a fuzzy finder search after typing query or initial fuzzy-search session
1 file = 1 keys 
2 files = 4 keys 
3 files= 8 keys
4 files = 13 keys

```

[...]

PR #165, opened 2015-12-17
Make shift-down open the selected file in the background

samrayner:
Prototype implementation of https://github.com/atom/fuzzy-finder/issues/70

Use shift-down to open the selected file in the background before advancing the list selection to the next item - useful for quickly opening multiple files from the finder.

This is almost certainly not a clean enough implementation and I have only ensured existing specs pass and not written additional ones. Hopefully it will let people play around with the functionality though and demonstrate its usefulness to be included in the package.


edwinramirez:
What ever happened to this????

PR #167, opened 2015-12-29
Async git

joshaber:
:warning: Dependent on https://github.com/atom/atom/pull/9213 :warning:

Use the new async git API.

I believe this is ready, but tests won't pass until there's a release of Atom including the async repo work.


joshaber:
:boom:

PR #174, opened 2016-02-16
Fix spec race

joshaber:
Refresh the entire status so we're totally up to date.


joshaber:
Background: our expected value was wrong because of the bugs fixed in https://github.com/atom/atom/pull/10827.

PR #180, opened 2016-03-09
Return project paths correctly if last-opened path is not in project

nathansobo:
screen shot 2016-03-08 at 7 20 40 pm

The previous code wasn’t returning `data` correctly if the last opened path was not found in the list, leading to the behavior pictured above. Basically if you switched away from an editor that was outside the project to a new editor, than toggled the fuzzy-finder, you'd see that screenshot.

/cc @joshaber for awareness but with :heart:.


joshaber:
Damn, good catch!

PR #181, opened 2016-03-16
Initialize File Finder with the text selected in active editor

gshah30:
Often, in the code we have file paths of files to be included or fetched. Opening such files currently requires one to copy the path and then paste it in the file finder search box.

This PR enables the file finder to detect text that is already selected in the active editor and search for it automatically on opening, thus saving the user an extra copy-paste operation.

Enabled as a configuration option.

settings1

This feature is available in WebStorm and was very handy. Had been missing it since I switched over from WebStorm to Atom.

Have been using it in Atom for 2 months now. Created the pull request today.

fuzzy-finder-enhancement

### DIFF
--- a/lib/fuzzy-finder-view.coffee
+++ b/lib/fuzzy-finder-view.coffee
@@ -39,7 +39,11 @@ class FuzzyFinderView extends SelectListView
 
   cancel: ->
     if atom.config.get('fuzzy-finder.preserveLastSearch')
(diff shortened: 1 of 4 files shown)


smashwilson:
I merged #314 which implements this yesterday, so this functionality will be included in the next release.

Thanks for giving this a shot :smile:

PR #194, opened 2016-04-18
Fix slow opening

### LINKED ISSUE #188: Panel takes 2-3 seconds to open
When I hit <kbd>Cmd</kbd>+<kbd>t</kbd> to open the fuzzy finder, it takes 2-3 seconds for the panel to open. This doesn't seem to be related to reindexing (that I can tell) because if I've just opened atom, it will start reindexing and show me that when I first open it. 

I've tested this using v1.7.0-beta3 (though this has been happening since v1.7.0-beta0) in safe mode and see the problem there as well so it doesn't seem to be a 3rd party package. When testing with v1.6.1, the panel opens near instantly.

It seems to be dependent on the size of the project (large rails project takes 2-3 seconds, project with 6 files is near instant).

Fuzzy Finder seemed like the right place for this, but if not let me know.

joshaber:
Fixes #188.

I don’t love how this turned out. The problem is that getting git status is async now, but the view needs to be created synchronously. I’m not sure if there’s a better way to accomplish this.

/cc @atom/core

### DIFF
--- a/lib/fuzzy-finder-view.coffee
+++ b/lib/fuzzy-finder-view.coffee
@@ -64,8 +64,7 @@ class FuzzyFinderView extends SelectListView
     @subscriptions?.dispose()
     @subscriptions = null
 
-  viewForItem: ({filePath, projectRelativePath, status}) ->
-
+  viewForItem: ({filePath, projectRelativePath}) ->
     # Style matched characters in search results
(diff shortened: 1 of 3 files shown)


BinaryMuse:
Thanks @joshaber, works great 👍

PR #198, opened 2016-04-26
add maxItem as option

shemerey:
Hi there, can we discuss this PR ? 

Sometimes it's nice to see all possible results what can fits on your screen. It's hard to override this hardcoded value. I think it's better to have it in config.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,13 @@
       "default": false,
       "description": "Remember the typed query when closing the fuzzy finder and use that as the starting query next time the fuzzy finder is opened."
     },
+    "maxFiles": {
+      "type": "integer",
+      "default": 10,
+      "minimum": 5,
+      "maximum": 20,
+      "description": "Maximum number of files to be displayed"
+    },
     "useAlternateScoring": {
       "type": "boolean",
       "default": true,
--- a/spec/fuzzy-finder-spec.coffee
+++ b/spec/fuzzy-finder-spec.coffee
@@ -21,6 +21,99 @@ rmrf = (_path) ->
   else
     fs.unlinkSync(_path)
 
+describe 'FuzzyFinder#maxItem configuration option', ->
+  [rootDir3] = []
+  [fuzzyFinder, projectView, workspaceElement, fixturesPath] = []
+
+  beforeEach ->
+    rootDir3 = fs.realpathSync(temp.mkdirSync('root-dir3'))
+
+    fixturesPath = atom.project.getPaths()[0]
+
+    wrench.copyDirSyncRecursive(
+      path.join(fixturesPath, "root-dir3"),
+      rootDir3,
+      forceDelete: true
+    )
+
+    atom.project.setPaths([rootDir3])
+
+    workspaceElement = atom.views.getView(atom.workspace)
+
+    waitsForPromise ->
+      atom.workspace.open(path.join(rootDir3, 'sample.js'))
+
+  dispatchCommand = (command) ->
+    atom.commands.dispatch(workspaceElement, "fuzzy-finder:#{command}")
+
+  waitForPathsToDisplay = (fuzzyFinderView) ->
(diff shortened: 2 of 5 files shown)


lee-dohm:
This setting won't change the size of the box that is drawn:

screen shot 2016-04-27 at 4 54 06 pm

It is currently hard coded to 10, but on my theme I only see 5. (I'm not sure if this is controlled by the theme or the package.)

shemerey:
@lee-dohm it will not by default, but if you want to customize how it looks, you have no options to control,   It make life harder for third party packages what consume `FF` as well. 

I my case I found it handy to open `FF` on full screen like so, but I can't see all results, and half screen is useful. 

FuzzyFinder with hardcoded value 

If I can control it by settings I can see something like this 

FuzzyFinder with settings

I understand that this is might be rare case but it looks like more pros than cons. can be wrong :)

lee-dohm:
So having a minimum of 5 and a maximum of 20 should be sufficient then, given your screenshot.

shemerey:
yes, I think so. I have no idea why I use 80 as max value :)

lee-dohm:
Sounds good. If you can make that change and add a test so that we don't inadvertently break this functionality in the future, I'd be happy to merge it.

shemerey:
@lee-dohm plz take a look, when you have time

package.json
```diff
@@ -39,6 +39,13 @@
       "default": false,
       "description": "Remember the typed query when closing the fuzzy finder and use that as the starting query next time the fuzzy finder is opened."
     },
+    "maxItem": {
```
50Wliu: Can this be renamed to `maxItems` instead?  That way when it's displayed it'll read as "Max Items" and not "Max Item".

In addition, perhaps items can be replaced with "files"?  So "maxFiles" and "Maximum number of files to be displayed"?

Just some :thought_balloon:s.

spec/fuzzy-finder-spec.coffee
```diff
@@ -1075,3 +1075,15 @@ describe 'FuzzyFinder', ->
 
           runs ->
             expect(projectView.list.find("li:contains(file.txt)")).toExist()
+
+      describe "maxItem config limitation", ->
```
lee-dohm: It's more important that the configuration is reflected in the view class than we do bounds checking in the specs. I'd like to see the following specs:

1. Instantiate the fuzzy finder view without changing the configuration and verify that the default is 10 as expected
1. Change the configuration (within the min and max values), instantiate a view and verify that the view has the right value
1. (Optional) Set the value to something out of bounds, instantiate a view and verify that the view has an in-bounds value

shemerey: thx for your comment, will do it as you described above

package.json
shemerey: fair enough, will fix it

shemerey:
@lee-dohm I've added some test according to your review. Since it's a package activation related option I have to introduce new top level describe with customised activation. 

plz take a look when you have time, and let me know if you have any thought about it

shemerey:
@lee-dohm, @50Wliu do you have any suggestions, or any thoughts about current PR ?

shemerey:
@lee-dohm, @50Wliu  ping

50Wliu:
Looks good for the most part.  My only issue with this is that the new spec was added at the top of the file in its own top-level describe block, when it should probably go somewhere under the existing `FuzzyFinder` describe  (such as https://github.com/atom/tree-view/blob/46076caaee988972175398e056e7adb14806b8cf/spec/tree-view-spec.coffee#L2456).

lee-dohm:
Thanks for the hard work and effort that you put in to this. I've reviewed it with the team and we've decided that we don't want to add this change. We're trying to control the complexity of Atom's core functionality to prevent costly regressions.

I apologize that the team didn't have this discussion sooner to save you the time you put into it.

Issue #111, opened 2015-06-29
On some folder, Cmd+T always tells me "Project is empty"


vincentisambart:
As you can see on the screenshot below:
![screen shot 2015-06-29 at 08 07 54](https://cloud.githubusercontent.com/assets/34704/8398733/0b060ac2-1e36-11e5-95a5-ef1c396f1a06.png)

Atom was started with `atom --safe <directoryname>`.
OS X 10.10.3, Atom 1.0.
Other folders work well.
(moved from atom/atom#7520)

vincentisambart:
It's probably because that directory is in the `.gitignore` of its parent directory.
Reproduction:
```
git clone --depth 1 https://github.com/CocoaPods/Rainforest.git
cd Rainforest
rake bootstrap # you can Ctrl+C as soon as the CocoaPods directory has been cloned
atom --safe CocoaPods
```
Then if you press Cmd+T you will get the screenshot above.

michaelaye:
Have the same problem and yes, I have another git repo above this project that ignores all folders to keep track of some general setting files that are applicable for all projects underneath. So my .gitignore has the line `*/`. Is there a workaround to still be able to use the fuzzy finder?

spencern:
Same issue here. 
Happening for directories that are `.gitignored` in parent for me as well.

maaarghk:
This is happening here also; I would expect the .gitignore which is up a few levels not to be read as the project itself has its own .git folder; is there a more appropriate place to raise this bug to do with the way ignoredNames are dealt with, or is this with the fuzzy-finder project?

maaarghk:
^ further to this, it seems fuzzy finder is the only thing which is affected, everything else manages alright. So I think this is the correct place for this bug.

aaronjensen:
From what I can tell, the issue only seems to be with `*/` being ignored. If you simply ignore `CocoaPods`  it works as expected.

maaarghk:
^ can confirm.

I have the following structure

```
/project
|
|-- things
|--|-- thing1
|--|-- thing2
|--|-- .gitignore
|
|-- .gitignore
```

If I put "things" in /project/.gitignore then fuzzy finder works as expected with thing1/thing2 folders as project base directories.

If I however put in /project/things/.gitignore :
```
*
!.gitignore
```
because I want the folder to remain empty in source control, then the fuzzy finder does not work correctly in the same situation.

kemayo:
[...]

PR #207, opened 2016-05-06
Don't ignore the whole project.

joshaber:
Fixes https://github.com/atom/fuzzy-finder/issues/111.

If the repository was embedded in another repository, then the inner repository itself could be ignored which meant we wouldn’t index anything.

This _might_ be less than ideal in that project roots which were previously entirely ignored now won’t be. But I gotta think that if you opened a project root, you don’t want the whole project ignored.

### DIFF
--- a/lib/load-paths-handler.coffee
+++ b/lib/load-paths-handler.coffee
@@ -46,7 +46,7 @@ class PathLoader
     @paths = []
 
   loadPath: (pathToLoad, done) ->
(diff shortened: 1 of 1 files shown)


foundryspatial-duncan:
For anyone still experiencing this problem, I just noticed the **php-integrator-base** package is overriding the fixed behavior and still causing the bug. Disabling the package has fixed it for me in 1.8.0

PR #215, opened 2016-05-17
Updated readme to explain jumping to line numbers.

calebmeyer:
If we'd like more explanation, I can add it. If we want another photo, I'll have to figure out how to get it on f.cloud.github.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,9 @@ Quickly find and open files using `cmd-t`.
   * `cmd-t` or `cmd-p` to open the file finder
   * `cmd-b` to open the list of open buffers
   * `cmd-shift-b` to open the list of Git modified and untracked files
+
+When opening a file, you can control the behavior.
+
   * `enter` defaults to opening the selected file without leaving the current pane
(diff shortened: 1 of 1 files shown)


lee-dohm:
If you can split out the <kbd>Enter</kbd>, <kbd>Shift+Enter</kbd> and line number stuff into a separate list, I think it would be more readable.

README.md
```diff
@@ -7,6 +7,7 @@ Quickly find and open files using `cmd-t`.
   * `cmd-shift-b` to open the list of Git modified and untracked files
   * `enter` defaults to opening the selected file without leaving the current pane
   * `shift-enter` defaults to switching to another pane if the file is already open there
+  * `:<line number>` will go directly to the line number you specify, or the last line if the number is larger
```
lee-dohm: Rewriting the beginning of this to:

> Adding `:<line number>` to the end of your search text will go directly ...

I think will be more clear.

biesmard:
I am the right person to develop

smashwilson:
:+1: Thanks!

PR #217, opened 2016-05-24
Windows compatible specs

damieng:


lib/project-view.coffee
```diff
@@ -81,7 +81,7 @@ class ProjectView extends FuzzyFinderView
         # throw ENOENT when it tries to get the realpath of all the project paths.
         # This catch block allows the file finder to still operate on the last
         # set of paths and still let the user know that something is wrong.
-        if error.code is 'ENOENT'
+        if error.code is 'ENOENT' or error.code is 'EPERM'
```
as-cii: Can we add a comment about this `EPERM` case, like we did for `ENOENT`? Also, what do you think about adding test coverage for this, to make sure we don't regress in the future?

as-cii:
One comment, otherwise it looks good! 👌

lib/project-view.coffee
damieng: Can definitely add a comment as to why it's there.  The test coverage is the fact it fails on Windows right now. I guess it would be possible to write a Mac test that also exercised the EPERM path but would seem to have less value once the Windows CI server runs specs?

Issue #158, opened 2015-11-27
Allow search with full path


eMerzh:
if your project is at /home/joe/project

you open the fuzzy-finder, and type "dir/myfile.js" it will display everything matching.

but if you are searching for "/home/joe/project/dir/myfile.js" , you won't find anything

it might be a good idea to allow absolute path search too...

wdyt?

PR #237, opened 2016-07-13
Add match for exact path ref #158

eMerzh:
Hi, 

i'm really new in the atom dev / coffee and all, 
but here is a small itch that i had using atom..
trying to past full path in the fuzzy finder and not having the ability to open the path directly...
might not be the place or the best way ....

please comment :)

### DIFF
--- a/lib/fuzzy-finder-view.coffee
+++ b/lib/fuzzy-finder-view.coffee
@@ -175,6 +175,9 @@ class FuzzyFinderView extends SelectListView
     else
       filteredItems = @items
 
+    if fs.isFileSync filterQuery
+      filteredItems.push(filePath: filterQuery, projectRelativePath: filterQuery)
+
     @list.empty()
     if filteredItems.length
       @setError(null)


50Wliu:
@eMerzh Can you please add some specs for this?  Thanks.

spec/fuzzy-finder-spec.coffee
```diff
@@ -585,6 +585,20 @@ describe 'FuzzyFinder', ->
           expect(atom.workspace.panelForItem(projectView).isVisible()).toBe false
           expect(inputView).toHaveFocus()
 
+  describe "when the filter text contains an absolute path", ->
+    it "it matches the given path", ->
```
50Wliu: No need for a double it.  `it "matches the given path"` is fine.

50Wliu:
Let's say I input `~/.atom/config.cson` as the filter query.  Will that show up?

spec/fuzzy-finder-spec.coffee
eMerzh: thanks for the review, 
not sure about your comment, you suggest to replace
```coffee
describe "when the filter text contains an absolute path", ->
   it "it matches the given path", ->
```
by 
```coffee
   it "it matches the given path", ->
```
?

eMerzh:
hum... nothing now...
but i think it might be a good addition ....
not sure how to do it though :/

spec/fuzzy-finder-spec.coffee
lee-dohm: He's suggesting replace:

```coffee
it "it matches the given path", ->
```

with

```coffee
it "matches the given path", -"
```

You don't need to use the word `it` twice in the same line :grinning:

eMerzh:
did i broke the tests? they all run ok on my machine

50Wliu:
Don't worry about the tests.

spec/fuzzy-finder-spec.coffee
```diff
@@ -585,6 +585,20 @@ describe 'FuzzyFinder', ->
           expect(atom.workspace.panelForItem(projectView).isVisible()).toBe false
           expect(inputView).toHaveFocus()
 
+  describe "when the filter text contains an absolute path", ->
```
50Wliu: Now that relative paths are supported, can you also add a test for that?

eMerzh: it's not really relative.... it's "normalized" , you can't do like ../current_file.html
although it could be interesting

but you cant do  
 ~/.atom/config.cson or /home/me/../hello

eMerzh:
is there something i can do to help this PR moving ?

Issue #156, opened 2015-11-26
Use a more realist debounce time


jeancroy:
This issue exist to helps issue such as  https://github.com/atom/fuzzy-finder/issues/153, as well as general responsiveness on slower computer or larger project. 

Currently fuzzy-finder inherit it's debounce delay from space-pen/select-list-view at [50ms](https://github.com/atom/atom-space-pen-views/blob/master/src/select-list-view.coffee#L21)

I understand we want instant result. A good definition of instant could be "as fast as one can type"
How fast do we type ? A quick google search tells me:

> The average person types between 38 and 40 words per minute -- between 190 and 200 characters per minute. (avg 5 char per word)

That is 3 characters per second.

> An average professional typist types usually in speeds of 50 to 80 wpm -- about 250-400 characters per minute. 

That is 4 - 6 characters per second.

---

50 ms delay = 20 characters per second **!**
150 ms delay = 6.66 characters per second
200 ms delay = 5 characters per second 
300 ms delay = 3.33 characters per second.

---

Fuzzaldrin have a 60k files benchmark.
It takes 150ms to run on a current year computer and about double on a 4 year computer.

Starting atom in my home folder, I have about 70k files.

---

The thing about the current delay is that it is VERY SHORT. 
Literally, it's very hard to trigger, unless we hold a key and let the OS repeat it.

I believe a good rule of thumb for debounce is to not send request faster than we can compute them.

Now, how fast we can compute those results ? The answer to that question really depend on the project and the machine, and because of that, I really prefers to [measure it](https://github.com/atom/autocomplete-plus/issues/541) rather than use a magic number.

With that being said, if a magic number it must be, a lot of things seems to point to just under 300ms.

lee-dohm:
> I understand we want instant result. A good definition of instant could be "as fast as one can type"

The industry accepted definition of "instant" [based on usability research](http://www.nngroup.com/articles/response-times-3-important-limits/) is 100ms.

Those figures on the average layman and average professional typist are sustained, long-form typing. When someone types in short bursts, especially from muscle memory ... they're going to be typing a lot faster.

jeancroy:
Fair point.

With that being said, the current behavior when user type faster than computation time is that the keyboard event get registered to a queue and played back later.

So a lag is going to be there anyways. The difference is one case have a 100% cpu lag, and discard computation as soon as it get done, and the other case wait idle or do other gui stuff while waiting.

Even 100ms that would be twice the room for computation.

[...]

PR #263, opened 2016-11-07
configurable input throttle

dansgithubuser:
Addresses https://github.com/atom/fuzzy-finder/issues/156

I'd like to write a spec that checks the actual output changing over time, but not sure how. And it could be argued that that particular behavior is SelectListView's responsibility.


lee-dohm:
Thanks for the effort you've invested in this. On discussing this with the team, we don't want to offer such an esoteric configuration value. Even if the current number is incorrect, we feel that there is a single number that is going to work for the overwhelming majority of cases. Also, we believe offering a configuration option for this will be confusing and prone to errors.

Thanks again for contributing! :bow:

PR #272, opened 2017-01-21
Save file names to file and load after restart

### LINKED ISSUE #271: Slow startup for large projects
Each time you open a project and bring up fuzzy-finder, it shows 'Indexing...' for several seconds. For large projects it's quite annoying.
Suggested solution: save per-project caches after reindex and load them on startup, delete old caches. I've implemented it [in my fork here](https://github.com/antelle/fuzzy-finder-with-faster-loading/blob/master/lib/project-view.coffee) and can make a pull request, if this idea is welcome.
On my project with ≈50k files, fuzzy finder can be used immediately after startup and it saves time.

antelle:
### Description of the Change

Hello, this is a quick fix for #271.

Here, I added a simple implementation of saving autocomplete data to a local file.
This way, we can load it after restart and show the ui immediately after a project is open.

The file list is deflated, because it contains a lot of similar strings and using the minimal compression level, we can get huge decrease in size.

### Alternate Designs

For now, this is the simplest implementation.

### Benefits

This change will bring better support for projects with a lot of files, e.g. >50k, removing quite a long wait time.
Old cache files are removed after a month, if there were no updates.

### Possible Drawbacks

A list of files will be saved to .atom folder, this may be unexpected. However old files will be removed in a month, if the corresponding projects are not opened frequently.

### Applicable Issues

#271

### DIFF
--- a/lib/project-view.coffee
+++ b/lib/project-view.coffee
@@ -37,7 +37,7 @@ class ProjectView extends FuzzyFinderView
       @reloadPaths = true
       @paths = null
 
-    if !@paths
+    if not @paths
       @tryLoadCachedProjectFiles()
 
     @cleanupOldFiles()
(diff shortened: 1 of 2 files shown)


antelle:
Guys, please take a look, I'd like to know your opinion about the idea and design. This PR lacks some tests, e.g. checking that we save only if there're changes, but they are quite fast to add, if the design looks ok from your point of view.

50Wliu:
I definitely think this could help. /cc @BinaryMuse

AndreasDahl:
What is the status of this PR?

maxbrunsfeld:
I don't think this is the right solution to this problem. The projects in your file could have changed since the last time you opened Atom, right?

Fundamentally, it should be possible to use an off-the-shelf tool like `ripgrep` for doing this indexing, eliminating the performance issue. We just haven't had time to switch the behavior. Until we do, I don't think it's worth compromising correctness for it.

PR #273, opened 2017-01-26
Replace atom-space-pen-views with atom-select-list

as-cii:
This pull request uses atom-select-list as a drop-in replacement for atom-space-pen-views. Feature-wise nothing should change, but this will help removing jQuery from Atom.

/cc: @ungb for 👀


ungb:
Looks good @as-cii I didn't see any issues while testing.

ungb:
I just tested on latest from your branch, I didn't find any regressions.  

One issue I noticed that also repros before your change is the setting `Search All Panes`.  

1. Disable Search All Panes
2. Click `cmd-b` to search from buffer
3.  **Expected:** only buffer from existing pane to show up. **Actual:** all results show up, if you click on a result, it is reopened in the active pane.

NOTE: if you have Search All Panes Enabled, it goes to the file in that pane.  

Search All Panes Disabled:
disabled



Search All Panes Enabled:
enable

NOTE: this issue I'm not sure classifies as a regression as it current exist, I can create another bug for it if needed.

PR #275, opened 2017-02-12
Add service provider

yuriihabrusiev:
### Description of the Change

This PR adds service provider, so other packages can toggle Fuzzy Finder with custom query text. Also it replaces changes from PR #137 and PR #181 can be implemented as a package.

### Alternate Designs

1. As introduced in #137.
2. Everyone make their own implementation of Fuzzy Finder (like in [related-files](https://github.com/raviraa/related-files) or [related](https://github.com/nick125/related))

### Benefits

Packages can toggle Fuzzy Finder with custom query.

### Possible Drawbacks

No drawbacks

### Applicable Issues

No issues

### To Do
- [x] Add service provider
- [x] Add specs
- [x] Add docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,3 +17,33 @@ This package uses both the `core.ignoredNames` and `fuzzy-finder.ignoredNames` c
 This package also will also not show Git ignored files when the `core.excludeVcsIgnoredPaths` is enabled.
 
 ![](https://f.cloud.github.com/assets/671378/2241456/100db6b8-9cd3-11e3-9b3a-569c6b50cc60.png)
+
+## API
+
+This package provides a service that allows you to trigger **Fuzzy Finder** from other Atom packages.
+
+### Consume service
+
+#### package.json
+```json
+{
+  "consumedServices": {
+    "atom.fuzzy-finder": {
+      "versions": {
+        "1.0.0": "consumeFuzzyFinder"
+      }
+    }
+  }
+}
+```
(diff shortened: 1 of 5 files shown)


yuriihabrusiev:
Hey guys, can anyone provide some feedback about this one. I'd really like to see this going forward.
cc @50Wliu, @as-cii, @lee-dohm

lee-dohm:
So this service would open the fuzzy finder with pre-populated search text? I'm not sure how beneficial this would be to other package authors. The two packages you mention provide a different set of files to select from and pre-populated search text wouldn't necessarily help with that.

yuriihabrusiev:
Hey @lee-dohm , thanks for your response!

Let me show you few cases:
- User wants to search files with name similar to word under cursor, selected text or quoted text. Like in Rails we have `<%= render 'path/to/partial' %>` and with this service we'll be able to find this partial file in one click. [I made little demo of this feature](https://github.com/yuriihabrusiev/go-to/blob/master/lib/go-to.js#L33).
- User wants to find files related to their current context. For example, in Rails if we are in model `Post`, with this service we'll be able find all files which start with `posts`.
- User wants to search for a file in some directory. Like pre-populate fuzzy-finder with `app/models` and search only there.

Sorry for Rails-only cases, but I'm sure many users will benefit from future like this. Also, Sublime Text has similar feature 😉.

Issue #269, opened 2017-01-08
responsiveness for large projects


dansgithubuser:
I often open my root directory in atom to get easy access to all the files I don't open often enough to open in their own project. This makes for a large index (300K files), and slow search times. For a full search, this is unavoidable, expected, and fine.

What's unnacceptable is that the slowdown is _on the UI thread_. I currently do not use fuzzy finder because of this.

If you're thinking of writing this off as innappropriate use, then fuzzy finder still needs to tell users not to attempt this and give them the UI thread back.

The obvious solution is to take the heavy operations off the UI thread, but there might be more to milk here. Just by going through fuzzy finder's issues, it seems there's a lot of clever heuristics that have been evolved over time. Maybe it's possible (natural?) to use these such that the most likely thing the user is looking for shows up quickly? Then we actually save the user a lot of time (on average), even when the entire operation would take a long time.

I originally approached this through https://github.com/atom/fuzzy-finder/issues/156 but chatting with @lee-dohm, he suggested I create a new issue that concentrates on the problem, not a possible solution.

dansgithubuser:
I wanted to attach a profile per lee's request, but when I save the file, an empty file comes out. I do not want to get off topic. I'll attach a screenshot. This is me typing asdf with slightly more than 50 ms between each letter.
![2017-01-08-033031_1366x768_scrot](https://cloud.githubusercontent.com/assets/1649921/21748460/d9e8b7ba-d553-11e6-97fe-774cf89f12f1.png)

driskell:
I'm experiencing this issue with a large repository I work with. It used to respond extremely quickly. I'm not sure exactly when it started but it now lags for well over a second between characters. Even using up/down key press to scroll through causes well over a second lag between movements (even though mouse wheel scrolling is instant.) It seems also that the project is reindexed every time I open the panel, even though nothing has changed. The label shows for a few seconds. Even waiting for the indexing to finish and then typing do I get the exact same problem of seconds of delays between characters.

Profile snapshot:
<img width="918" alt="screen shot 2017-04-25 at 20 35 14" src="https://cloud.githubusercontent.com/assets/939815/25404182/d9481290-29f6-11e7-9bcd-abeb3ae7db9d.png">

lee-dohm:
@driskell Is the repository you're working with public so that we can use it to reproduce the problem?

driskell:
Unfortunately, it's private. I just inspected and the item list is approximately 89000 items.

[...]

PR #281, opened 2017-03-09
Consume File-Icons' element-icon service

Alhadis:
This PR applies to the `tabs` package what's been described in [`atom/tree-view#1146`](https://github.com/atom/tree-view/pull/1146) for the `tree-view` package. The PR's updated wording describes this scenario best, so I'll just copy+paste:

-------------------

I realised too late that your icon-service fulfils a different purpose from what our package needed. [We wrote our own instead](https://github.com/file-icons/atom#integration-with-other-packages), which is built on a dubious but (currently) stable foundation of monkey-patches.

Opinions differ on the subject, but mine is that monkey-patching code you didn't write is bad, and should **only** be used as a last resort (DOM polyfills notwithstanding). Which is precisely what I've [been forced to do](https://github.com/file-icons/atom/tree/master/lib/consumers) to get dynamic icon-assignment working.

This needs to be fixed at a formal level, because:

1. **Patches don't persist if File-Icons is deactivated and reactivated.**  
I've actually chosen not to fix this, because it would step outside the expected lifecycle of the package. E.g., when a user deactivates a package, they expect it to leave no traces in the workspace. I also can't imagine this happening *too* often, but still...

2. **Any innocent change to your packages could break stuff.**  
The obvious danger of monkey-patching what wasn't expected or supposed to be changed. We have specs to alert us of breakage, but there's no telling what would be involved in the repair. We both know this is the wrong approach.

Now, I don't know how the Atom team would feel about adding support for third-party package services. Ideally, this would be addressed on the level of Atom's core icon-API. However, the differences of our needs (as well as the specific use-case of our needs) make me hesitant to propose a change to your existing service, especially because @as-cii has stressed he prefers keeping its functionality simple for the time being. In light of that, it would be more appropriate and ergonomic to support a third-party service in the interim, should a mutually-compatible solution be realised in future.

What this service does
=================

The `file-icons.element-icons` service, when consumed, provides a function to add dynamic icons to DOM elements. The function is to be called by the package on any element that's supposed to represent a filesystem resource (either files, or directories):

```js
let fileIcon = document.querySelector("li.file-entry > span.icon");
addIconToElement(fileIcon, "/path/to/file.txt");
```

Calling the function returns a Disposable that clears the icon-node instance from memory, which should be done once the view is destroyed.

Note there's no requirement to specify whether a path is a file or directory. Our [heavy-duty filesystem API](https://github.com/file-icons/atom/tree/master/lib/filesystem) takes care of the heavy lifting... I've even plans to separate it from File-Icons in the form of a standalone Node module, so other package authors can benefit from my hard work too. :)

Related pull-requests
=================
* [`atom/tree-view#1146`](https://github.com/atom/tree-view/pull/1146)
* [`atom/tabs#412`](https://github.com/atom/tabs/pull/412)
* [`atom/find-and-replace#870`](https://github.com/atom/find-and-replace/pull/870)
* **`atom/fuzzy-finder#281`**
* [`atom/archive-view#47`](https://github.com/atom/archive-view/pull/47)


Alhadis:
I can imagine how busy the Atom team are right now smoothing over bugs with the recent docked-views feature, as well as concentrating on fixing a [pretty severe memory issue](https://github.com/abe33/atom-pigments/issues/308) in the 1.18 beta.

That being said, I really don't wanna come off sounding naggy or impatient... but could somebody give me an update on this? The rotten monkey-patches I wrote have been allowed to exist for exactly half-a-year now (at least in my timezone, which officially ticked over to June over an hour ago....)

/cc @lee-dohm / @50Wliu / @as-cii

lee-dohm:
Thanks for the ping @Alhadis ... I'm going to get some :eyes: on this :+1:

lee-dohm:
@Alhadis we've added this (and its related PRs) to the list for prioritization. I noticed that file-icons was mentioned as a potential performance issue in https://github.com/atom/fuzzy-finder/issues/269. Would you mind doing some testing to see if this change helps or hurts performance in that scenario?

Alhadis:
Certainly! :D Which settings do you prefer I use when recording the timeline snapshot? I tend to get them all confused... 😞

Alhadis:
Atom hanged when it was trying to save the recorded CPU profile, but that's what it usually does when I have my machine's root or home directory opened as a project... :( I managed to screen-cap the first profile, which was recording performance *without* this PR's changes in effect:

screen shot 2017-06-09 at 3 51 17 am

screen shot 2017-06-09 at 3 52 15 am

screen shot 2017-06-09 at 3 53 49 am

lee-dohm:
@Alhadis if you can manage to save the profile, zip it up and email it to me at lee-dohm@github.com? I believe saving the profile saves all the data and then I can fiddle with it on my end :+1:

Thanks very much for helping out :bow:

Alhadis:
I already tried, but it saved a blank file. :( As a matter of fact, this is what it always gives me whenever I try saving the CPU profile... :(

50Wliu:
You'll need to test on Atom 1.19.0 in order to get a non-blank profile.

lee-dohm:
What @50Wliu said :grinning: https://github.com/atom/atom/issues/13767

Alhadis:
Ugh... good grief. 😞 

Should I profile the other packages while I'm running a HEAD-build of Atom v1.19, then...?

lee-dohm:
Yes, please :pray:

Alhadis:
Damn. Opening my user directory as an Atom project already shows how bad an issue this is... I'll e-mail you the first CPU profile I captured. It was done using dev-mode and *no* forked packages.

Please let me know if you have difficulties extracting it: it's a 23.3 MBs file I had to 7-Zip.

Alhadis:
Profiling the other packages hasn't really revealed much of a difference. The hit to performance is mostly at startup, when `file-icons` has to load and hack certain files.

Alhadis:
Conflicts fixed, and the bitter taste of CoffeeScript washed away.

Alhadis:
The 'hell's this...?

~~~
standard.cmd : standard: Use JavaScript Standard Style (https://standardjs.com)
At line:158 char:13
+             & "$standardpath" lib/**/*.js
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (standard: Use J...standardjs.com):String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
standard: Run `standard --fix` to automatically fix some problems.
  C:\projects\fuzzy-finder\lib\icon-services.js:44:76: Extra semicolon.
~~~

Yet there were no lint errors locally:

~~~console
λ FuzzyFinder (file-icons): ./node_modules/.bin/standard .
λ FuzzyFinder (file-icons):
~~~

nathansobo:
I appreciate your patience and dedication to making Atom better!

nathansobo:
Published as 1.7.0 and upgraded in atom/atom@d035e41f378497cc2d23b9f7983adebc1bd1820e.

I should have asked you to write some basic tests for this. Would you be willing to do a follow up PR that adds some basic tests so we don't break this in the future? I'm going to hold off on the other 3 PRs until there's some test coverage in place.

Thanks @Alhadis.

nathansobo:
Whoops, I had to revert the upgrade commit because there's an issue in with v8 snapshot generation. I'll take a look.

nathansobo:
Okay, I fixed the snapshot issue in 2c1fe65802860aa6c85a891e0ed58f390f493f8f and re-upgraded to 1.7.1 in atom/atom@058a42f7c2ff1a5ac35578b9b28ef32862f0d158. Basically, you can't interact with the `atom` require at eval-time in snapshotted code, and constructing the `IconServices` object at eval time was causing that to happen. I replaced the eval-time construction with a lazy getter function.

Alhadis:
Oh, you guys are using `mksnapshot`. That's clever. Should the other PRs be amended too?

> I should have asked you to write some basic tests for this. Would you be willing to do a follow up PR that adds some basic tests so we don't break this in the future?

Ah, I never notice that. Sure thing!

nathansobo:
> Should the other PRs be amended too?

Yeah, maybe you already got to it, but we need to avoid eval-time references to the `atom` require or any built-in node modules in any package that's going to be bundled.

Alhadis:
What about the complementary `DefaultFileIcons` class? That's being exported as a singleton instance in much the same way. Should that use lazy getters too?

Alhadis:
Think I'll leave it for now. `DefaultFileIcons` feels like it should be merged into the `IconServices` singleton more than anything else. I'd prefer not to make any more changes this late, however.

PR #293, opened 2017-05-02
Don't wait for items to be updated before showing fuzzy finder

### LINKED ISSUE #288: fuzzy-finder doesn't withhold keyboard events that fire during re-index and panel opening
### Prerequisites

 * [x] Reproduced the problem in Safe Mode
 * [x] Checked that your issue isn't already filed

### Description

Since `1.16.0` either the project reindexing perf drastically downgraded, or there's a regression (or both) that causes the text typed after user hits `ctrl + p` being added to editor instead of the fuzzy-finder panel, provided user starts to type _before_ reindexing is done and panel opens.

AFAIK this never happened to me before and I always start to type well before the panel opens (waiting for and ensuring the panel is open before typing is really bad UX).

This causes major problems if one's not careful and the nonsense text goes through linting, CRs etc. (not that often -- but still causes major pain due to constant refactor, backtracking etc.)

### Steps to Reproduce

1. open a project with moderate amount of files so that reindexing takes considerable time (0.5-1s+)
2. blur & refocus atom to [ensure reindexing](https://github.com/atom/fuzzy-finder/issues/88)
3. `ctrl + p` and start typing right away

**Expected behavior:** should enqueue the keyboard events until the panel opens (or whatever other implem), and seamlessly paste whatever user typed before it opened

**Actual behavior:**  the text that's typed _before_ the panel opens is entered into file buffer

**Reproduces how often:** 100% (well, practically every time I fuzzy-search for file after de-focusing atom, which is nearly always)

### Versions

atom `1.16.0`, fuzzy-finder `1.5.0`, win 7

### related

- [#188 - Panel takes 2-3 seconds to open](https://github.com/atom/fuzzy-finder/issues/188)
- [#88 - Fuzzy finder reindex whole project everytime](https://github.com/atom/fuzzy-finder/issues/88)

If I checked correctly, between atom `1.15.0` → `1.16.0`, one related bump happened:

- `fuzzy-finder` between `1.4.1` → `1.5.0`

as-cii:
Fixes #288.

Previously, we were waiting until the items had been populated before showing the fuzzy finder. However, this was causing an odd UX experience that made users accidentally type query characters into already opened buffers.

With this pull request, on toggle, we will show the fuzzy finder synchronously and only then request an animation frame where we populate the items.

/cc: @atom/maintainers

### DIFF
--- a/lib/buffer-view.js
+++ b/lib/buffer-view.js
@@ -24,9 +24,9 @@ export default class BufferView extends FuzzyFinderView {
       })
 
       const paths = Array.from(new Set(editors.map((editor) => editor.getPath())))
-      await this.setItems(paths)
       if (paths.length > 0) {
         this.show()
+        await this.setItems(paths)
       }
     }
   }
--- a/lib/git-status-view.js
+++ b/lib/git-status-view.js
@@ -23,8 +23,8 @@ export default class GitStatusView extends FuzzyFinderView {
 
         }
       }
-      await this.setItems(paths)
       this.show()
+      await this.setItems(paths)
     }
   }
 
--- a/lib/project-view.js
+++ b/lib/project-view.js
@@ -49,8 +49,8 @@ export default class ProjectView extends FuzzyFinderView {
     if (this.panel && this.panel.isVisible()) {
       this.cancel()
     } else {
-      await this.populate()
       this.show()
+      await this.populate()
     }
   }
 


Ben3eeE:
I was able to consistently reproduce the issue by opening the fuzzy-finder right after launching Atom. I used a folder located on a network share that is slow.
I tested 7 times with and without this PR. Without it reproduced 7/7 times (100%) and with this it didn't reproduce.
Both have the same delay before the view is shown for me.

I'll keep this linked to keep testing 👍 🚢

as-cii:
Thanks, Linus! ❤️

dwelle:
Has this already landed? It seems so, because the bug doesn't happen to me after editor loses focus, anymore. But it still happens after clean atom start (and possibly in few other cases that I can't repro).

Ben3eeE:
It landed in `1.17.0-beta5`. I tested both with a clean start of Atom and with the window losing focus and was not able to reproduce this bug.

dwelle:
Huh.. I'm on stable, and it _did_ seem weird that there was no `patch` bump, so there's that. But no idea why it started working well for me for the atom defocus/focus case. Before, it never opened before re-index, while now it always does, so I thought the patch must have already landed.

Ben3eeE:
@dwelle I had the same thing happen to me where defocus/focus case suddenly started working :grinning: It might have been something external that was slowing down Atom enough to cause this problem for me.  Since this PR I have not reproduced it in either case and the defocus case did happen fairly regularly for me.

Thank you for updating that you are on stable :bowing_man: This has not been fixed in Atom 1.16 so it's expected you still experience this issue if it is also an issue on Atom 1.16. Let us know if you experience problems when using Atom 1.17 :zap:

PR #305, opened 2017-07-08
Support :row:column in fuzzy-finder-view

mehcode:
### Description of the Change

Add support for a second colon to indicate the column to jump to.

For instance: `sample.js:4:10`

### Alternate Designs

None that I can think of.

### Benefits

We could bind the `go-to-line:toggle` command to this package and insert a `:`. This would remove a package from needing to exist clearing up a touch of maintenance burden.

### Possible Drawbacks

None that I can think of.

### Applicable Issues

None.

### DIFF
--- a/lib/fuzzy-finder-view.js
+++ b/lib/fuzzy-finder-view.js
@@ -44,7 +44,7 @@ export default class FuzzyFinderView {
           this.previousQueryWasLineJump = true
           this.selectListView.update({
             items: [],
-            emptyMessage: 'Jump to line in active editor'
+            emptyMessage: 'Enter a <row> or <row>:<column> to go there. Examples: "3" for row 3 or "2:7" for row 2 and column 7'
           })
         } else if (this.previousQueryWasLineJump && !isLineJump){
           this.previousQueryWasLineJump = false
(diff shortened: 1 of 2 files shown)


lib/fuzzy-finder-view.js
```diff
@@ -44,7 +44,7 @@ export default class FuzzyFinderView {
           this.previousQueryWasLineJump = true
           this.selectListView.update({
             items: [],
-            emptyMessage: 'Jump to line in active editor'
+            emptyMessage: 'Enter a <row> or <row>:<column> to go there. Examples: "3" for row 3 or "2:7" for row 2 and column 7'
```
50Wliu: This message seems overly verbose.  What about just `Jump to line:column in active editor`?

mehcode: I just copied it from the existing go-to-line package as I was intending to follow this PR with changes to `ctrl-g` to use a command from this package so as to deprecate the `go-to-line` package.

I agree that's a terribly long message. 

Should I still change it?

50Wliu: Hmm...I'm not that sure.  I'll try to get back to you after discussing whether or not we should deprecate go-to-line.

Issue #161, opened 2015-12-10
Fuzzy find and show results from selection?


wonderbeyond:
![image](https://cloud.githubusercontent.com/assets/228361/11704975/4bda2606-9f27-11e5-8e4c-7de0487787f7.png)
Is there an easy way to fuzzy find files from the selection?

lee-dohm:
Not currently, no. Though this sequence doesn't seem terribly complicated:

1. Copy keybinding
1. Fuzzy finder keybinding
1. Paste keybinding
1. <kbd>Enter</kbd>

stowball:
I'm so used to this feature from Brackets. It's a shame Fuzzy Finder doesn't support it as it reduces 2 keystrokes. Efficiency++ :D

danielsiwiec:
I'd love to see this too

JosephDAM:
Hi @lee-dohm , first of all ty for the fuzzy-finder. For my question, I just wanted to know if this is in development, is already implemented, or if it is goin to be at some point.

Regards

50Wliu:
@JosephDAM This feature is currently not in development, though we would accept a well-crafted PR adding this functionality.

ColCh:
Just to notify subscribed people, I've created PR for this feature: https://github.com/atom/fuzzy-finder/pull/314

It works for all type of fuzzy-finders - `Find File`, `Find Buffer`, `Find Modified File`

stowball:
Can't wait to try it out, https://github.com/ColCh. Thanks a lot!

PR #314, opened 2017-08-27
Prefill query from selection

ColCh:
### Description of the Change

When you select text, then activate fuzzy-finder, it will look for selection in active editor.

If there is some text selected, it will use it as query - and select text inside of query as well

### Alternate Designs

Didn't noticed any. Only in Nuclide's `nuclide-quick-open` there is analogue for it, but you can't use it without installing entire `nuclide` package

### Benefits

Now you don't need to Copy-Paste your selection in fuzzy finder.

This speedups programming performance :)

### Possible Drawbacks

May be opened a bit* slowly, because it looks for editor selected text

* didn't benchmarked... Also, didn't noticed any performance drop on my machine (MBP 2015)

### Applicable Issues

resolves https://github.com/atom/fuzzy-finder/issues/161


lee-dohm:
We are interested in taking this feature. We need two modifications first:

1. We need this functionality behind a configuration option
1. We need to default to the current behavior (so default to off)

Once that's done, we'll take a deeper look.

Thanks for the submission!

ColCh:
Done.

Now this behaviour is turned on by config variable.
And this config variable is turned *off* by default, meaning current behaviour.
Also, I've merged master changer with decaffeinated filed.

Please review, ping @lee-dohm

smashwilson:
:+1: Works like a charm. Thanks for the contribution! :tada:

ColCh:
@smashwilson great! Thank you!

PR #336, opened 2018-02-17
Keep existing selection after reindexing

### LINKED ISSUE #210: Jump to first item after reindexing
This is really more of an annoyance than a bug:
After opening the fuzzy finder and entering a search string, the item selection jumps to the start of the list as soon as the 'Reindexing project' message is gone.

1. Is it really reindexing the whole project everytime?
2. It leads to opening the wrong file quite often - which is very annoying.

50Wliu:
### Requirements

### Description of the Change

After a reindex completes, all the items in the select list get regenerated.  If someone had a file selected, the refresh would cause the selection to jump back to the first element.  This PR fixes that by manually re-selecting the previous file after the reindex completes.

### Alternate Designs

I hope there's an alternate design; this one didn't come out that pretty.

### Benefits

Selection will not unexpectedly change after a reindex completes.

### Possible Drawbacks

The for loop I added creates some delay, especially for larger projects since it loops through every single file.

### Applicable Issues

Fixes #210

/cc @Ben3eeE

### DIFF
--- a/lib/fuzzy-finder-view.js
+++ b/lib/fuzzy-finder-view.js
@@ -292,12 +292,25 @@ export default class FuzzyFinderView {
     }
   }
 
-  setItems (filePaths) {
+  async setItems (filePaths) {
     this.items = this.projectRelativePathsForFilePaths(filePaths)
     if (this.isQueryALineJump()) {
       return this.selectListView.update({items: [], loadingMessage: null, loadingBadge: null})
     } else {
-      return this.selectListView.update({items: this.items, loadingMessage: null, loadingBadge: null})
+      let selectedItem = this.selectListView.getSelectedItem()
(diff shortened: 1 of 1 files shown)


Ben3eeE:
@50Wliu Thanks for opening this ⚡️ LGTM. I couldn't notice any delay even with a larger project and slow computer where I noticed the issue this fixes!

🚢 if you are fine with the drawback!

nfarina:
This issue has been driving me crazy - would love to have this fixed! Any reason it can't be merged? Seems like a slam dunk—straightforward code and small scope of changes.

50Wliu:
Forgot about this PR.  Let me revisit it in the next few days and see if I can remember my logic here.

PR #347, opened 2018-05-16
Fix flaky test re "toggling when the project has multiple paths" 🤞

jasonrudolph:
Prior to this change, I was able to reliably reproduce the failures described in  https://github.com/atom/atom/issues/17325 when running `apm test` locally with Atom 1.29.0-dev-3e4b386. After this change, the tests reliably pass for me. Hopefully this resolves the flakiness described in https://github.com/atom/atom/issues/17325. 😅

_Most_ of the tests in `spec/fuzzy-finder-spec.js` use `async/await`, but the flaky tests were using a _mix_ of `async/await` and `waitsFor`. This change updates the flaky tests to only use `async/await`.

### DIFF
--- a/spec/fuzzy-finder-spec.js
+++ b/spec/fuzzy-finder-spec.js
@@ -254,13 +254,13 @@ describe('FuzzyFinder', () => {
         describe('socket files on #darwin or #linux', () => {
           let socketServer, socketPath
 
-          beforeEach(() => {
+          beforeEach(async () => {
             socketServer = net.createServer(() => {})
             socketPath = path.join(rootDir1, 'some.sock')
-            waitsFor(done => socketServer.listen(socketPath, done))
+            await socketServer.listen(socketPath)
           })
 
-          afterEach(() => waitsFor(done => socketServer.close(done)))
+          afterEach(async () => await socketServer.close())
 
           it('ignores them', async () => {
             await projectView.toggle()


spec/fuzzy-finder-spec.js
```diff
@@ -254,13 +254,13 @@ describe('FuzzyFinder', () => {
         describe('socket files on #darwin or #linux', () => {
           let socketServer, socketPath
 
-          beforeEach(() => {
+          beforeEach(async () => {
             socketServer = net.createServer(() => {})
             socketPath = path.join(rootDir1, 'some.sock')
-            waitsFor(done => socketServer.listen(socketPath, done))
+            await socketServer.listen(socketPath)
```
maxbrunsfeld: Looking at [the docs for `Server.listen`](https://nodejs.org/api/net.html#net_server_listen), it looks like the method just returns the `Server` itself, and not a `Promise`. So I would think that using `await` on the return value would be a noop.

It doesn't look like [`close`](https://nodejs.org/api/net.html#net_server_close_callback) returns a promise either, so `await` should be a noop there too.

I would think that you'd have to pass a callback to both methods; something like:

```js
await new Promise(resolve => socketServer.listen(socketPath, resolve))

// ...

await new Promise(resolve => socketServer.close(resolve))

```

It's interesting that this fixes the tests though. Maybe the listening and closing are so fast that they already complete before the next block executes?

PR #349, opened 2018-05-27
prevent current editor from showing up first

jacekkopecky:
### Requirements

### Description of the Change

The current open editor can end up being the first result in fuzzy search. This change makes sure it isn't.

I have a project with HTML and JS files named the same but the file extension, e.g. `project-marking.html` and `project-marking.js`, and I often want to quickly open the other one using fuzzy finder. However, fuzzy finder puts the `.js` file first, so if I'm there, open fuzzy finder, type `projmar`, pressing enter would just take me where I already am.

I would like to open fuzzy finder, type `projmar`, press enter, and be in **the other file that I know matches**.

This PR is a quick hack, an RFC really, that implements this change. Every time we filter, it checks that if we have more than one match, the first one shouldn't be the current open buffer – if it is, it switches the first two matches.

### Alternate Designs

I suspect that we might be able to cache the current open buffer's path when opening the fuzzy finder, but I don't know enough about the code to quickly guess where would be appropriate.

Also, this kind of sorting decision might be moved to fuzzaldrin, with some kind of *value that shouldn't be put first* parameter, but I don't have the time right now to investigate.

### Benefits

This allows me quicker to change files with fuzzy finder.

### Possible Drawbacks

If somebody already has in their muscle memory that they open fuzzy finder, type a query, and do *down* and *enter*, this would potentially give them the wrong file. 

### Applicable Issues

None I could find.

### DIFF
--- a/lib/fuzzy-finder-view.js
+++ b/lib/fuzzy-finder-view.js
@@ -122,12 +122,12 @@ class FuzzyFinderView {
                 if (filtered.length > 1) {
                   const activePaneItem = atom.workspace.getActivePaneItem()
                   if (activePaneItem && activePaneItem.getPath && activePaneItem.getPath() === filtered[0].filePath) {
-                    const x = filtered[0];
-                    filtered[0] = filtered[1];
-                    filtered[1] = x;
+                    const x = filtered[0]
+                    filtered[0] = filtered[1]
+                    filtered[1] = x
                   }
                 }
-                return filtered;
+                return filtered
               }
 
               return items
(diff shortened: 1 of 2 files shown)


jacekkopecky:
Sorry about the lint, I didn't notice standard.js was in use. The latest travis build failure seems to be something intermittent about the last few tests (not the one I added).

thomasjo:
> The latest travis build failure seems to be something intermittent about the last few tests (not the one I added).

Yeah, I restarted the build and it's :green_heart: now.

lee-dohm:
Thanks very much for the contribution.

Upon reviewing this change with the team, we've decided that this poses too many risks with regard to people's established workflows. We can recommend that this could be useful as a community package that is more custom-built for this proposed workflow. Because of this, we're going to close the PR.

Thanks again for your hard work and interest in Atom :+1:

jacekkopecky:
No worries.

🎉 welcome new Atom package: [fuzzy-finder-switch-first](https://atom.io/packages/fuzzy-finder-switch-first)

PR #354, opened 2018-09-17
Close tab from fuzzy-finder list view

oxilumin:
### Description of the Change

A new command to kill a tab is added.
As long as the tab being closed is not the last tab on the list - move selection to the next tab. If tab is the last tab - move selection to the new "last" tab after removal to avoid selection-jumping in long lists.
If the tab being closed is the last tab - close it and the popup.

### Benefits

When working in mouse-less setup it's not very convenient to close tabs by activating them and then closing each one separately. This basically copies feature from sublime text

### Possible Drawbacks

None

### Applicable Issues

None

### DIFF
--- a/keymaps/fuzzy-finder.cson
+++ b/keymaps/fuzzy-finder.cson
@@ -18,3 +18,4 @@
 
 '.fuzzy-finder atom-text-editor[mini]':
   'shift-enter': 'fuzzy-finder:invert-confirm'
+  'ctrl-w': 'fuzzy-finder:kill-tab'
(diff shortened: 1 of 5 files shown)


rsese:
Thanks @oxilumin!

Can you add tests? As mentioned in the `Requirements` section, we require tests to guard against regressions so tests are necessary before the team can review.

Also:

> Applicable Issues
> None

This section is referring to an open issue that your change is fixing/implementing - is there an open issue for this?

oxilumin:
Hi @rsese, I've switched from atom couple month ago because of slow performance and lack of features.

Issue #358, opened 2018-10-31
Misleading caption displayed during line number entry


jtagscherer:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When a user wants to jump to a line in the currently opened file, they can use the fuzzy finder to do so by entering a colon followed by the line number. If the line number becomes invalid after further input, the finder warns the user. This warning persists even if the user corrects their input, potentially misleading them. Confirming the dialog and jumping to the entered line number works regardless of the alert.

### Steps to Reproduce

1. Start Atom and open an arbitrary file.
2. Press <kbd>⌘</kbd>+<kbd>T</kbd> to open the fuzzy finder.
3. Enter a colon and a valid line number, for example `:42`. The dialog hints that this will make the editor jump to the corresponding line.
4. Make the entered line number invalid, for example by appending a non-numeric character to your input. The dialog changes its hint, making the user aware that their input is invalid.
5. Make the entered line number valid again, for example by deleting the non-numeric character.

![Visual representation of the steps](https://user-images.githubusercontent.com/3198913/47791645-72d12f00-dd1a-11e8-8228-506fb8818eb6.gif)

**Expected behavior:** The dialog should return to displaying the previous hint about jumping to the entered line in the active editor.

**Actual behavior:** The dialog keeps displaying the alert until the search box is cleared, although the input is in fact valid.

**Reproduces how often:** Everytime.

### Versions

```
Atom    : 1.32.0
Electron: 2.0.9
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

```
apm  2.1.2
npm  6.2.0
node 8.9.3 x64
atom 1.32.0
python 3.6.4
git 2.17.2
```

OS: macOS Mojave (10.14)

PR #359, opened 2018-10-31
Correctly update the caption during line number entry

jtagscherer:
### Description of the Change

As outlined in #358, the status message of the fuzzy finder did not update correctly while entering line numbers. This was due to an issue within the logic used to update the status message: The alert was not replaced by the standard message when the input became valid again because the `previousQueryWasLineJump` flag would stay true during the process, preventing the necessary block of code to be run.

The changes within this pull request fix this issue while simplifying the code in question a bit.

This is my first contribution to Atom, and I hope I did everything right. Do point me towards any mistakes I might have made though.

### Alternate Designs

N/A

### Benefits

The fuzzy finder always displays an appropriate status message when entering line numbers, no longer misleading users.

### Possible Drawbacks

N/A

### Applicable Issues

#358

### DIFF
--- a/lib/fuzzy-finder-view.js
+++ b/lib/fuzzy-finder-view.js
@@ -43,21 +43,22 @@ class FuzzyFinderView {
           this.iconDisposables = null
         }
         const isLineJump = this.isQueryALineJump()
-        if (isLineJump && /\D/.test(this.selectListView.getQuery().substr(1))) {
+        if (isLineJump) {
           this.previousQueryWasLineJump = true
-          this.selectListView.update({
-            items: [],
-            emptyMessage: null,
-            errorMessage: 'Invalid line number'
-          })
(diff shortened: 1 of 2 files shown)


50Wliu:
I've restarted the build and it looks good now.

jasonrudolph:
I tried this change out locally and verified that it resolves the problem described in #358.

I reviewed the diff, and it looks like a solid improvement to me.

I made some minor tweaks to the test in https://github.com/atom/fuzzy-finder/compare/94f251281297211fea91c4c6149d56567be8fefc...d53f278abff1887e80ab33fabe7e806579150cc1.

Thanks for this contribution, @jtagscherer! :bow:

jtagscherer:
Happy to help! Thank you for your improvements to my PR, @jasonrudolph.

PR #364, opened 2019-01-27
Provide accurate error messages when jumping to columns

### LINKED ISSUE #363: Wrong hint when jumping to the column of a line
### Description

Since https://github.com/atom/fuzzy-finder/pull/360, the fuzzy finder allows users to jump not only to lines within files but also to columns within lines by using a second colon. Since https://github.com/atom/fuzzy-finder/pull/359, the hint in the finder's caption is correctly updated while jumping to lines. I implemented both of these pull requests simultaneously, and since I had no control over when they were merged, I unfortunately could not include jumping to columns within the hints.

In consequence, the hint in the finder view erroneously considers input with two colons invalid. This creates a bad user experience by harming the discoverability of the feature of jumping to columns. Not only does the finder not hint at the feature, but it actively discourages users from using the feature by displaying an error message.

bug

### Steps to Reproduce

1. Open any file with some content
2. Open the fuzzy finder, e.g. by pressing <kbd>CMD</kbd>+<kbd>P</kbd>
3. Insert a line and column separated by colons, e.g. `:4:2`

**Expected behavior:**

When entering a line and a column, the finder should display an appropriate hint, e.g. "Jump to position in active editor". If the part before the second colon is invalid, it should display "Invalid line number". If the part after the second colon is invalid, it should display "Invalid column number".

**Actual behavior:**

The hint does not differentiate between jumping to lines and columns. Even worse, it displays an error message when trying to jump to a column, even though the input is technically valid.

**Reproduces how often:**

Always.

### Versions

**Atom Version:**

```
Atom    : 1.34.0
Electron: 2.0.16
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

**APM Version:**

```
apm  2.1.3
npm  6.2.0
node 8.9.3 x64
atom 1.34.0
python 3.6.4
git 2.17.2
```

**Operating System:**

macOS Mojave (10.14.2)

50Wliu:
### Requirements

### Description of the Change

With #360, you can now specify a specific column when jumping to a line. The error reporting for invalid line numbers, however, was not updated to allow columns as well. This PR adds additional cases to the error handling such that it reports column errors correctly and does not inadvertently display "Invalid line number" when attempting to navigate to a column.

### Alternate Designs

None.

### Benefits

An incorrect error is removed and there is better discoverability of the feature.

### Possible Drawbacks

None.

### Applicable Issues

Fixes #363, /cc @jtagscherer

### DIFF
--- a/lib/fuzzy-finder-view.js
+++ b/lib/fuzzy-finder-view.js
@@ -45,7 +45,20 @@ class FuzzyFinderView {
         const isLineJump = this.isQueryALineJump()
         if (isLineJump) {
           this.previousQueryWasLineJump = true
-          if (/\D/.test(this.selectListView.getQuery().substr(1))) {
+          const query = this.selectListView.getQuery()
+          if (/^:\d+:\d*\D/.test(query)) {
+            this.selectListView.update({
+              items: [],
+              emptyMessage: null,
+              errorMessage: 'Invalid column number'
+            })
+          } else if (/^:\d+:/.test(query)) {
+            this.selectListView.update({
+              items: [],
+              emptyMessage: 'Jump to line and column in active editor',
+              errorMessage: null
+            })
+          } else if (/^:\d*\D/.test(query)) {
             this.selectListView.update({
               items: [],
               emptyMessage: null,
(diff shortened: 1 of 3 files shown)


rafeca:
Looks great! You can add an assertion to check that the message is correct around the following lines:

https://github.com/atom/fuzzy-finder/blob/94b91a7fb2a5d9f9eaf6d2d17e71eb5420c9acf3/spec/fuzzy-finder-spec.js#L916-L920

rafeca:
I've just added some tests, we can merge this now! 🎉

50Wliu:
Thanks @rafeca! Was a bit busy so I wasn't able to get around to finishing this.

PR #313, opened 2017-08-22
Preview selected files in list #5

neilthawani:
This PR is a clone of @jwuwork's contribution in https://github.com/atom/fuzzy-finder/pull/113 with the merge issues addressed.

Copy/paste of their PR details:
Ok folks, you asked for it. Here is the new feature for previewing selected files.

To enable this feature, go into fuzzy-finder Settings and check the new "Preview Selection" option:

image

Obviously, this works best if you have "Use Preview Tabs" option turned on from the tab package Settings.

Issue #182, opened 2016-03-18
Feature Enhancement: Ability to open files in pending view from the fuzzy-finder search bar


christopherciufo:
**Current and suggested behavior**

There is currently no way to open a file as a pending item from the fuzzy search. I suggest adding the ability for files that are being searched in the fuzzy finder search to open in pending mode. From my understanding, there are 3 possible logical approaches to implementing this without breaking the current configuration.

**Options**

1. Have a boolean option either in *Core Settings* or in the `fuzzy-finder` package settings that would cause all files opened from the fuzzy search to be in pending mode. This would be exactly like pre-1.6.0 behavior when the feature was in the `tabs` package.

2. Have a command option that allows keybindings to map to opening a fuzzy search that will always open files in pending mode. (Ex: I can map `cmd-t` to open a finder that will open files normally, and map `cmd-p` to open a finder that will open files as a pending item).

3. Implement a feature similar to *Search All Panes* in the `fuzzy-finder` package that allows some sort of on-the-fly choice of opening the file that is in focus on the fuzzy finder search to be opened as a pending item. (Ex: `enter` on the current highlighted file will open it normally, but something like `cmd-enter` might open it in pending mode).

If implemented carefully, I believe options 1 & 2, or 1 & 3 can be implemented in harmony. Options 2 & 3 probably won't work together.

**Why would the enhancement be useful to most users?**

I imagine that there are many users, such as myself, that do not use the tree view for most file opening operations, and prefer to use the command palette. I almost exclusively open files through `cmd-p` or `cmd-t`, and found the old *Preview Tabs* feature helpful since I would normally be browsing to check if the current file is the one I want to keep open. Option 1 above would return the behavior to this state, but I think options 2 or 3 are a better solution since they allow for more flexibility. I believe most people would love the ability to have the choice between opening their files as a pending or standard pane item.

@lee-dohm @kuychaco I would be willing to implement any of these if you guys think one in particular is the better direction.

**Atom Version:** 1.6.0+
**OS and Version:** All platforms & supported versions

kuychaco:
Thanks for the great writeup @christopherciufo. I agree with your assessments and think it would be great to have this as an enhancement to the pending view feature. 

I say go with Option 3. It gives users the option of making the decision of whether to open in pending/non-pending mode at the last second. Whereas with Option 2, users have to make the decision up front and if they change their mind it's an extra keystroke to switch to opening fuzzy-finder in the other mode. 

[...]

PR #113, opened 2015-07-04
Preview selected files in list #5

jwuwork:
Ok folks, you asked for it.  Here is the new feature for previewing selected files.

To enable this feature, go into **fuzzy-finder** Settings and check the new "Preview Selection" option:

screen shot 2015-07-03 at 10 25 34 pm

Obviously, this works best if you have "Use Preview Tabs" option turned on from the **tab** package Settings.

PR #368, opened 2019-03-10
Preview selected files in list #5 #182

pheen:
### Description of the Change

It has been brought up in a few issues and PRs (#113, #313) that having a Sublime Text like preview in the fuzzy finder would be helpful. This implementation should behave identical to Sublime's finder preview.

`didChangeSelection` gives a callback for user query changes in the fuzzy finder, and fortunately after that Atom's existing pending panes can handle the hard part.

### Alternate Designs

`didChangeSelection` is a little awkward to use because it's called more often than exactly when a user makes a change. For example when initially opening the fuzzy finder it can be called twice, so `this.showNextPreview` is needed to skip those events. I think a better solution would involve changes to `SelectListView` which is probably more than this issue warrants.

### Benefits

Allow possibly faster/easier decision making when selecting files, especially when searching for similar items. Also to make Sublime Text users happy :)

### Possible Drawbacks

There should be no impact when the feature is disabled. I haven't noticed any issues so far when enabled.

### Applicable Issues

#5, #182

Mar-10-2019 03-44-40

### DIFF
--- a/lib/fuzzy-finder-view.js
+++ b/lib/fuzzy-finder-view.js
@@ -30,7 +30,7 @@ class FuzzyFinderView {
 
         return query
       },
-      didCancelSelection: () => { this.cancel() },
+      didCancelSelection: () => { this.cancel({ destroyPreview: true }) },
       didConfirmSelection: (item) => {
         this.confirm(item, {searchAllPanes: atom.config.get('fuzzy-finder.searchAllPanes')})
       },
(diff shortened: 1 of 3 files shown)


lee-dohm:
Thanks for opening this PR and for the hard work you put into it.

Upon reviewing with the team today, we've decided that we don't want to own this functionality because of the various performance and stability impacts that it could have on the fuzzy-finder package. For example, what if you type some characters that match the data files in a particular directory that are 5GB each? For this reason, we're not going to accept this PR and we're closing the related issues.

However, we are going to open an issue for someone to create a [provided service](https://flight-manual.atom.io/behind-atom/sections/interacting-with-other-packages-via-services/) that could allow other community packages to integrate with fuzzy-finder to provide this behavior. We would be open to reviewing an [RFC](https://github.com/atom/atom/blob/master/docs/rfcs/000-template.md) for a design of this proposed service in atom/fuzzy-finder. If you or anyone is interested in working on this, please open a new issue for the RFC for review by the maintainer team.

Thanks again for your hard work and wanting to make Atom better.

Issue #371, opened 2019-03-19
Log the time it takes to index the project


rafeca:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

In order to understand better the real world performance of the `fuzzy-finder` and allow us to take better decisions, we want to start logging the time it takes to re-index a project ([code pointer](https://github.com/atom/fuzzy-finder/blob/master/lib/path-loader.js#L15-L23)).

Some additional information that we want to log along the time:

* **Number of files in the repository**: the time will vary a lot depending on the project size so we need to be able to filter results by similar project sizes.
* **is ripgrep enabled**: Our hypothesis is that `ripgrep` will have an important impact on the indexing time, so we want to log if it's enabled to be able to compare results.

rafeca:
This has already landed `atom@master`

PR #372, opened 2019-03-21
Start logging the time to index a project

rafeca:
### Description of the Change

This PR makes use of the `atom/metrics` package to be able to start logging the time it takes to index a project by the fuzzy finder.

The format of the data that's being logged is the following:

* **eventType**: `fuzzy-finder-v1`
* **metadata**

  | field | value |
  |-------|-------|
  | `ec` | `time-to-crawl`
  | `el` | Crawler type (`ripgrep` or `fs`)
  | `ev` | Number of crawled files

Sample payload

```json
{
  "eventType": "fuzzy-finder-v1",
  "durationInMilliseconds": 539,
  "metadata": {
    "ec": "time-to-crawl",
    "el": "ripgrep",
    "ev": 49,
    "cd2": "x64",
    "cd3": "x64",
    "cm1": 180,
    "cm2": 83,
    "sr": "1680x1050",
    "vp": "1680x591",
    "aiid": "dev"
  },
  "date": "2019-03-21T17:05:32.706Z"
}
```

### Test plan

- [x] Check that `atom/telemetry` sends the correct payload to GitHub's backend. To do so, I've had to edit the code of the `atom/telemetry` package (more specifically, I've manually modified `node_modules/telemetry-github/out/index.js` to set the variable `this.isDevMode` to `false` ([code](https://github.com/atom/telemetry/blob/master/src/index.ts#L108)). I've also had to increase how often Atom sends metrics, since by default it only sends them every 24 hours. To do so, I've set the variables `DailyStatsReportIntervalInMs` and `ReportingLoopIntervalInMs` located in the previous file to something like `15000` ([code](https://github.com/atom/telemetry/blob/master/src/index.ts#L20-L23)).
- [x] Check that the data gets propagated correctly to our analytics pipeline.

### Alternate Designs

Do not log this?

### Benefits

Knowing the time it takes to crawl the project by the fuzzy finder on real users will help us decide how much effort should be put on improving its performance.

### Possible Drawbacks

We're logging a little bit more data from users.

### Applicable Issues

https://github.com/atom/fuzzy-finder/issues/371

/cc @telliott27

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -1,0 +1,24 @@
+# Events specification
+
+This document specifies all the data (along with the format) which gets send from the Fuzzy Finder package to the GitHub analytics pipeline. This document follows the same format and nomenclature than the [Atom Core Events spec](https://github.com/atom/metrics/blob/master/docs/events.md).
+
+## Counters
+
+Currently the Fuzzy finder does not log any counter event.
+
+## Timing events
+
+#### Time to crawl the project
+
+* **eventType**: `fuzzy-finder-v1`
+* **metadata**
+
+  | field | value |
+  |-------|-------|
+  | `ec` | `time-to-crawl`
+  | `el` | Crawler type (`ripgrep` or `fs`)
+  | `ev` | Number of crawled files
+
+## Standard events
+
+Currently the Fuzzy Finder does not log any standard event.
--- a/docs/events.md
+++ b/docs/events.md
@@ -1,10 +1,10 @@
 # Events specification
 
-This document specifies all the data (along with the format) which gets send from the Fuzzy Finder package to the GitHub analytics pipeline. This document follows the same format and nomenclature than the [Atom Core Events spec](https://github.com/atom/metrics/blob/master/docs/events.md).
+This document specifies all the data (along with the format) which gets sent from the Fuzzy Finder package to the GitHub analytics pipeline. This document follows the same format and nomenclature as the [Atom Core Events spec](https://github.com/atom/metrics/blob/master/docs/events.md).
 
 ## Counters
 
-Currently the Fuzzy finder does not log any counter event.
+Currently the Fuzzy finder does not log any counter events.
 
 ## Timing events
 
@@ -21,4 +21,4 @@ Currently the Fuzzy finder does not log any counter event.
 
 ## Standard events
 
-Currently the Fuzzy Finder does not log any standard event.
+Currently the Fuzzy Finder does not log any standard events.
--- a/lib/reporter-proxy.js
+++ b/lib/reporter-proxy.js
@@ -1,0 +1,35 @@
+
+module.exports = class ReporterProxy {
+  constructor () {
+    this.reporter = null
+    this.queue = []
+
+    this.eventType = 'fuzzy-finder-v1'
+  }
+
+  setReporter (reporter) {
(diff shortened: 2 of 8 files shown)


rafeca:
One limitation of this implementation is that if the user opts out from metrics, the metrics reporter service won't initialize and the queue that keeps the events that haven't been sent will grow indefinitely...

This issue is already happening in the `welcome` package ([code](https://github.com/atom/welcome/blob/master/lib/reporter-proxy.js)), while the `github` package has a fix for it by adding [quite a bit of logic](https://github.com/atom/github/blob/master/lib/reporter-proxy.js).

IMHO this should not be the responsibility of every package that wants to consume the `metrics` service: packages shouldn't care about whether users have or not opted in into metrics and ideally the metrics service should always return a `reporter` which will log or not depending on the user preferences.

I can implement this solution as part of `atom/metrics` if people think that this is the right approach 😃

docs/events.md
```diff
@@ -0,0 +1,24 @@
+# Events specification
+
+This document specifies all the data (along with the format) which gets send from the Fuzzy Finder package to the GitHub analytics pipeline. This document follows the same format and nomenclature than the [Atom Core Events spec](https://github.com/atom/metrics/blob/master/docs/events.md).
```
jasonrudolph:
```suggestion
This document specifies all the data (along with the format) which gets sent from the Fuzzy Finder package to the GitHub analytics pipeline. This document follows the same format and nomenclature as the [Atom Core Events spec](https://github.com/atom/metrics/blob/master/docs/events.md).
```

docs/events.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Counters
+
+Currently the Fuzzy finder does not log any counter event.
```
jasonrudolph:
```suggestion
Currently the Fuzzy finder does not log any counter events.
```

docs/events.md
```diff
@@ -0,0 +1,24 @@
...
+
+  | field | value |
+  |-------|-------|
+  | `ec` | `time-to-crawl`
+  | `el` | Crawler type (`ripgrep` or `fs`)
+  | `ev` | Number of crawled files
+
+## Standard events
+
+Currently the Fuzzy Finder does not log any standard event.
```
jasonrudolph:
```suggestion
Currently the Fuzzy Finder does not log any standard events.
```

lib/reporter-proxy.js
```diff
@@ -0,0 +1,35 @@
+
```
jasonrudolph: ☝️ Can we remove this extra line?

jasonrudolph:
> Sample payload

Thanks for including the sample payload. Seeing this sample made it much easier for me to get a quick feel for these changes. :zap:

> IMHO this should not be the responsibility of every package that wants to consume the `metrics` service: packages shouldn't care about whether users have or not opted in into metrics and ideally the metrics service should always return a `reporter` which will log or not depending on the user preferences.

I agree. :+1:

rafeca:
> Can you describe how you'll verify that these changes are having the desired effect?

I've added some extra information in the PR summary 😄 

> Would it make sense to add a test or two for this new functionality?

Yes! I just added a couple of tests (and they actually made me find an issue when queueing events 😅 ).

Issue #370, opened 2019-03-19
Improve performance when filtering large projects


rafeca:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

`fuzzy-finder` is super laggy when filtering results for large projects. The UI gets blocked and the user experience is not great.

![bad](https://user-images.githubusercontent.com/408035/54600680-a0a2e000-4a3d-11e9-8790-0487558b0381.gif)

### Steps to Reproduce

1. Open a medium to large repo from Atom (large repo: https://github.com/mozilla/gecko-dev).
2. Open the fuzzy finder and start searching for some text.
3. Realize the lag.

**Expected behavior:** the fuzzy finder UI should either update faster or we should add some kind of debouncing behaviour for large projects

**Actual behavior:** The filtering seems to be happening on each key stroke, and they get queued so when the user stops typing the filtering of previous key strokes is still being processed and it takes some time to "empty the queue".

**Reproduces how often:** Always for large projects.

Aerijo:
> Expected behavior: the fuzzy finder UI should either update faster or we should add some kind of debouncing behaviour for large projects

Would it be possible to instead / also run it on a different thread (or whatever gets executed independently of the UI)? Much like how VS Code extensions cannot directly impact the core editor's performance.

rafeca:
> Would it be possible to instead / also run it on a different thread (or whatever gets executed independently of the UI)? Much like how VS Code extensions cannot directly impact the core editor's performance.

Yup, running as many things as possible outside of the UI thread is the ideal solution (in fact this is how the `fuzzy-finder` crawls the files). For this specific scenario there are two things that we need to consider though:

- If the filtering gets run on a different thread, we'll have to add logic to handle/limit the concurrency: the filtering would become an async operation so if you start typing `c` + `co` + `con` we have to make sure that if the filtering process for `co` finishes after the one for `con` we don't display the wrong results.
[...]

PR #373, opened 2019-03-28
Do not call `setItems()` every time that the fuzzy finder is shown

rafeca:
### Description of the Change

This PR unblocks the work to use `native-fuzzy` to improve the filtering performance by mitigating one of [its concerns](https://github.com/atom/fuzzy-finder/issues/370#issuecomment-477246644).

Basically, before this PR the UI was blocked by the execution of the `setItems()` method every time the fuzzy finder was shown (and it delayed its appearance in the UI). This was not a big problem before because the `setItems()` method used to be quite fast, but with the addition of `native-fuzzy` we need to do more expensive things here (we need to serialize and pass the whole list of items to the `native-fuzzy`).

### Alternate Designs

N/A

### Benefits

* Unblocking a big performance improvement.
* Less work done every time the fuzzy finder is opened

### Possible Drawbacks

N/A

### Applicable Issues

https://github.com/atom/fuzzy-finder/issues/370

### DIFF
--- a/spec/fuzzy-finder-spec.js
+++ b/spec/fuzzy-finder-spec.js
@@ -71,6 +71,12 @@ describe('FuzzyFinder', () => {
     return conditionPromise(() => fuzzyFinderView.element.querySelectorAll('li').length > 0)
   }
 
+  async function waitForReCrawlerToFinish (fuzzyFinderView) {
+    return conditionPromise(
+      () => !fuzzyFinderView.element.querySelector('.loading .loading-message')
+    )
+  }
+
   function eachFilePath (dirPaths, fn) {
     for (let dirPath of dirPaths) {
       wrench.readdirSyncRecursive(dirPath).filter((filePath) => {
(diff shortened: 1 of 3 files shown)


spec/fuzzy-finder-spec.js
```diff
@@ -734,6 +747,9 @@ describe('FuzzyFinder', () => {
           await projectView.toggle()
 
           expect(PathLoader.startTask).toHaveBeenCalled()
+
+          await waitForReCrawlerToFinish(projectView)
```
rafeca: There's a very small behaviour difference with this PR: before when adding or removing a project from the current Atom window the fuzzy finder was temporarily shown as empty while it was recrawling the new list of projects. Now, it displays the old results while the re-crawling happens (which is consistent with what happens when a new file is added/removed).

Before:

before

Now:

now

Based on its description, this test was not doing the assertion correctly: instead of checking that the end list of results was empty was just checking that the temporary loading window was empty (which also could cause some flakiness if the crawling happened faster than the time it takes to open the fuzzy finder), so I fixed it.

PR #374, opened 2019-03-28
Use `fuzzy-native` as an alternate scoring system

rafeca:
### Description of the Change

This PR does two things:
* Renames the `useAlternateScoring` config option to convert it to an enum called `scoringSystems` (using the same default value as before).
* Adds a new `scoringSystem` called `fast` (🤷‍♂️) which uses `fuzzy-native` to dramatically improve filtering times on large projects.

### Alternate Designs

A few other options have been considered:
* Adding a debouncer without changing the current filtering logic.
* Keep the current filtering logic but move it to an async Task.
* Use the [`nuclide-prebuilt-libs`](https://github.com/facebook-atom/nuclide-prebuilt-libs) package instead of `fuzzy-native`.
* Create a new package with the `fuzzy-native` filtering logic executed in a background thread.

At the end, this solution is what IMHO gives more benefits and less tradeoffs per unit of effort 😅

For more information about benefits/tradeoffs, check out [the discussion on the applicable issue](https://github.com/atom/fuzzy-finder/issues/370).

### Benefits

Drastically faster filtering for large projects, which are now usable.

Type of project | master | `native-fuzzy` | Perf improvements
---- | --- | --- | --- |
Small | 10-24ms | 1ms | 👍 **17X faster**
Medium | 25-160ms | 1.5-9ms | 👍 **17X faster**
Large | 400-1800ms | 20-73ms |  👍 **22X faster**

For the benchmark I've configured `native-fuzzy` to use a single thread. When using 4 threads the time to filter improves by ~2-3X on large projects but stays almost the same for medium/small.

demo

### Possible Drawbacks

Yet a new scoring system added. As a follow-up we may decided to remove the `standard` scoring system, which was superseeded by the `alternate` one 3 years ago. 

### Applicable Issues

https://github.com/atom/fuzzy-finder/issues/370

### DIFF
--- a/lib/fuzzy-finder-view.js
+++ b/lib/fuzzy-finder-view.js
@@ -2,6 +2,8 @@ const {Point, CompositeDisposable} = require('atom')
 const fs = require('fs-plus')
 const fuzzaldrin = require('fuzzaldrin')
 const fuzzaldrinPlus = require('fuzzaldrin-plus')
+const NativeFuzzy = require('fuzzy-native')
+
 const path = require('path')
 const SelectListView = require('atom-select-list')
 
@@ -15,16 +17,19 @@ const getIconServices = require('./get-icon-services')
 const SCORING_SYSTEMS = {
   STANDARD: 'standard',
   ALTERNATE: 'alternate',
+  FAST: 'fast'
 }
 
-module.exports =
-class FuzzyFinderView {
+const MAX_RESULTS = 10
+
+module.exports = class FuzzyFinderView {
   constructor () {
     this.previousQueryWasLineJump = false
     this.items = []
+
     this.selectListView = new SelectListView({
       items: this.items,
-      maxResults: 10,
+      maxResults: MAX_RESULTS,
       emptyMessage: this.getEmptyMessage(),
       filterKeyForItem: (item) => item.label,
       filterQuery: (query) => {
@@ -90,6 +95,14 @@ class FuzzyFinderView {
           case SCORING_SYSTEMS.STANDARD:
             matches = fuzzaldrin.match(label, filterQuery)
             break
+          case SCORING_SYSTEMS.FAST:
+            this.nativeFuzzyForResults.setCandidates([label])
+            const items = this.nativeFuzzyForResults.match(
+              filterQuery,
+              {maxResults: 1, recordMatchIndexes: true}
+            )
+            matches = items.length ? items[0].matchIndexes : []
+            break
           default:
             matches = fuzzaldrinPlus.match(label, filterQuery)
             break
@@ -141,6 +154,27 @@ class FuzzyFinderView {
 
         if (this.scoringSystem === SCORING_SYSTEMS.STANDARD) {
           this.selectListView.update({filter: null})
+        } else if (this.scoringSystem === SCORING_SYSTEMS.FAST) {
+          if (!this.nativeFuzzy) {
+            this.nativeFuzzy = new NativeFuzzy.Matcher(this.items.map(el => el.label))
+
+            // We need a separate instance of the fuzzy finder to calculate the
+            // matched paths only for the returned results. This speeds up considerably
+            // the filtering of items.
+            this.nativeFuzzyForResults = new NativeFuzzy.Matcher([])
+          }
+
+          this.selectListView.update({
+            filter: (items, query) => {
+              if (!query) {
+                return items
+              }
+
+              return this.nativeFuzzy.match(query, {maxResults: MAX_RESULTS}).map(
+                result => this.convertPathToSelectViewObject(this.getAbsolutePath(result.value))
+              )
+            }
+          })
         } else {
           this.selectListView.update({
             filter: (items, query) => {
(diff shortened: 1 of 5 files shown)


lib/fuzzy-finder-view.js
```diff
@@ -118,16 +149,38 @@ class FuzzyFinderView {
...
+          }
+
           this.selectListView.update({
             filter: (items, query) => {
-              return query ? fuzzaldrinPlus.filter(items, query, {key: 'label'}) : items
+              if (!query) {
+                return items
+              }
+
+              return this.nativeFuzzy.match(query, {maxResults: MAX_RESULTS}).map(
```
rafeca: For now, we're executing `native-fuzzy` on a single thread, but we can easily change this in the future

lib/fuzzy-finder-view.js
```diff
@@ -299,21 +358,49 @@ class FuzzyFinderView {
...
+    const [rootPath, projectRelativePath] = atom.project.relativizePath(filePath)
+    const label =
+      rootPath && projectHasMultipleDirectories
+        ? path.join(path.basename(rootPath), projectRelativePath)
+        : projectRelativePath
+
+    return {uri: filePath, filePath, label}
+  }
+
+  getAbsolutePath (relativePath) {
```
rafeca: This method is not great because it'll do a few fs accesses via `fs.existSync`, but it should not be very expensive because it does it against the list of filtered items (which is maximum `10` items).

This means that it'll do `10 * number of opened projects` file system accesses. Since I don't expect users having many projects opened in the same window, this should be fine

spec/fuzzy-finder-spec.js
```diff
@@ -93,11 +93,19 @@ describe('FuzzyFinder', () => {
     }
   }
 
-  for (const useRipGrep of [false, true]) {
-    describe(`file-finder behavior (ripgrep=${useRipGrep})`, () => {
+  const testPermutations = [
+    [false, 'standard'],
+    [true, 'standard'],
+    [false, 'alternate']
+    [false, 'fast']
```
rafeca: I don't think it's worth to test on every single permutation of `useRipGrep:scoringSystem`, since these two params are orthogonal, so I've just selected a few permutations that ensure that both variables get fully tested.

rafeca:
uhm, `appveyor` shows failures due to `python2` not being available on they vms... Investigating

rafeca:
Ok, I've been able to build and use `fuzzy-native` locally on a Windows machine, but I found two things:

* `npm install` on `fuzzy-finder` does not generate the correct binary on `windows`. I had to run `apm install` to be able to generate the correct one (for electron).
* The package is not prepared to read binaries generated for electron 2.0 (more info: https://github.com/hansonw/fuzzy-native/blob/master/lib/main.js#L10-L11).

I think that both issues are related with the fact that the package is not prepared for electron v2.0.

I'm not sure if this is related to the issue that's happening on AppVeyor (I don't think so).

50Wliu:
> * `npm install` on `fuzzy-finder` does not generate the correct binary on `windows`. I had to run `apm install` to be able to generate the correct one (for electron).

This is expected; `apm install` should _always_ be used for installing Atom packages.

50Wliu:
Try adding `image: Visual Studio 2015` to `appveyor.yml`

rafeca:
> This is expected; apm install should always be used for installing Atom packages.

Oh interesting, I still have to learn a lot about Atom 😅 

> Try adding image: Visual Studio 2015 to appveyor.yml

I'm gonna try, thanks!

rafeca:
Ok, that fixed the issue 😄 

Now the remaining issue is the one caused by these lines: https://github.com/hansonw/fuzzy-native/blob/master/lib/main.js#L10-L11

rafeca:
Ok, using a fork that adds support for electron v2 we're able to run the tests (only that 2 of them fail 😅), I'm gonna try to reproduce the failures on my windows machine (on OSX all of them pass)

We're making progress!

rafeca:
Yay! AppVeyor builds are finally passing 😄 There was a test that didn't contain correct Windows paths and it didn't work well with the path transformations that the logic for `fuzzy-native` does 😅

rafeca:
For some reason Travis builds get queued forever: https://travis-ci.org/atom/fuzzy-finder/builds/512728958?utm_source=github_status&utm_medium=notification ... I'll check back tomorrow

package.json
```diff
@@ -11,6 +11,7 @@
     "fs-plus": "^3.0.0",
     "fuzzaldrin": "^2.0",
     "fuzzaldrin-plus": "^0.6.0",
+    "fuzzy-native": "git://github.com/rafeca/fuzzy-native.git#electron-v2",
```
maxbrunsfeld: Maybe you should publish this to NPM as `@atom/fuzzy-native` or something. That's what we've done for a few forks that we've had in the past, like `nsfw` and `react`.

rafeca: Sounds good, I'll do that

rafeca:
I've forked `fuzzy-native` to https://github.com/atom/fuzzy-native and created a `@atom/fuzzy-native` package with v0.7.0 that just contains the fix for electron v2.

It seems that `npm` caching logic is not super smart and the package cannot be found yet from CI systems... I'll wait some time and retrigger the build.

I'll also set up a similar CI system that the original `fuzzy-native` package has (again, I cannot do it now since Travis cannot find the new GitHub project yet)

rafeca:
Ok, after a few retries all tests passing, I'm proceeding to merge this PR 🙏

PR #375, opened 2019-03-29
Add metrics for fuzzy-finder filtering time

rafeca:
### Summary

In order to understand better the real world performance of the `fuzzy-finder` and allow us to take better decisions, we want to start logging the time it takes to execute a filtering action.

Along with the time taken to do the filtering, we want to log some additional metadata:

* **Number of files in the repository**: the time will vary a lot depending on the project size so we need to be able to filter results by similar project sizes.
* **Scoring system**: Our hypothesis is that the new `fast` scoring system implemented [here](https://github.com/atom/fuzzy-finder/pull/374) will have an important impact on the filtering time, so we want to log which is the scoring type that's being used.

Since the `standard` scoring system is not controlled by the fuzzy finder (but via `atom-select-list`) we cannot easily log it from here. That's not a big deal since we're planning on deprecating that system.

### Schema of the data

* **eventType**: `fuzzy-finder-v1`
* **metadata**

  | field | value |
  |-------|-------|
  | `ec` | `time-to-filter`
  | `el` | Scoring type (`alternate` or `fast`)
  | `ev` | Number of files in the project

Sample JSON payload

```json
{
  "eventType": "fuzzy-finder-v1",
  "durationInMilliseconds": 539,
  "metadata": {
    "ec": "time-to-filter",
    "el": "fast",
    "ev": 789,
    "cd2": "x64",
    "cd3": "x64",
    "cm1": 180,
    "cm2": 83,
    "sr": "1680x1050",
    "vp": "1680x591",
    "aiid": "dev"
  },
  "date": "2019-03-21T17:05:32.706Z"
}
```

### Verification process

- [x] Check that Atom is sending the expected JSON payload to the GitHub backend systems by enabling the metrics debug mode.
- [x] Check that the the events get stored on our pipelines correctly.

//cc @telliott27

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -19,6 +19,17 @@ Currently the Fuzzy finder does not log any counter events.
   | `el` | Crawler type (`ripgrep` or `fs`)
   | `ev` | Number of crawled files
 
+#### Time to filter results
+
+* **eventType**: `fuzzy-finder-v1`
+* **metadata**
+
+  | field | value |
+  |-------|-------|
+  | `ec` | `time-to-filter`
+  | `el` | Scoring system (`alternate` or `fast`)
+  | `ev` | Number of items in the list
+
 ## Standard events
 
 Currently the Fuzzy Finder does not log any standard events.
--- a/lib/fuzzy-finder-view.js
+++ b/lib/fuzzy-finder-view.js
@@ -23,10 +23,11 @@ const SCORING_SYSTEMS = {
 const MAX_RESULTS = 10
 
 module.exports = class FuzzyFinderView {
-  constructor () {
+  constructor (metricsReporter) {
     this.previousQueryWasLineJump = false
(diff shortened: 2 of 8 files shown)


rafeca:
Something worth to notice is that this is going to create an event for every single keypress on the fuzzy finder. I'm going to add some logic to throttle the number of events so we send maximum one e.g every minute or so.

docs/events.md
```diff
@@ -22,3 +22,18 @@ Currently the Fuzzy finder does not log any counter events.
 ## Standard events
 
 Currently the Fuzzy Finder does not log any standard events.
```
jasonrudolph: I think we want to remove this `## Standard Events` section here, since it's included at the bottom of this file.

rafeca: ups, my mistake... copy pasta 😅

Issue #376, opened 2019-04-01
Show prompt to enable "fast mode" on large projects


rafeca:
### Context

Atom 1.37 will ship with two important performance improvements ([[1]](https://github.com/atom/fuzzy-finder/pull/374) and [[2]](https://github.com/atom/fuzzy-finder/pull/369)) on the fuzzy finder, both of them behind config flags and disabled by default (we want to minimize risks of breaking users' workflows by having both of them disabled for a version).

Under the hypothesis that not many people will enable these features if we don't do anything,
we want to show a prompt to a subset of users so they can easily enable them and we can get some early data and potential issues before we enable the feature by default.

### Constrains

These are the constrains that our solution will have:

* It should be shown to users that will benefit the most. from these features. So the benefits received will overweight the potential issues.
* It should have the less amount of friction possible. A user should do the minimum amount of work possible to enable these features.
* We should not spam users. If a user chooses to not opt in, we shouldn't ask him again.
* It should be clear how to disable the features. If something goes wrong, users should know how to go back to the normal fuzzy finder.

### Potential solution

Show a [`Notification`](https://atom.io/docs/api/v1.35.1/Notification) to the user if the crawling process took more than e.g 5s and if the notification hadn't been shown before.

The notification should have clear call to action buttons that don't require the user to navigate to other parts of the application. Example of notification:

<img width="475" alt="Screen Shot 2019-04-01 at 12 35 25" src="https://user-images.githubusercontent.com/408035/55321746-b15e4780-547a-11e9-904f-884e082954fa.png">

### Metrics

In order to understand how successful is this notification, we can log the following data:

* Number of times that the notification was shown.
* Number of times that people enabled fast mode.
* Number of times that people discarded the notification.

rafeca:
I have been pairing with @jasonrudolph early today on this issue and we've pivoted quite a bit from the original idea and improved the whole solution for the prompt.

This is what the new prompt is going to look like:

<img width="597" alt="Screen Shot 2019-04-03 at 19 54 49" src="https://user-images.githubusercontent.com/408035/55501321-76127300-564a-11e9-9c7c-12584c5b04e0.png">

* To make it easier for users to disable the fast mode or report issues (this was something hard to do with the previous solution), we're going to move the prompt directly to the fuzzy finder. There, we're going to show a message that informs the user about the "fast mode".
* Once the fast mode is enabled, the prompt will change to allow users to either disable it or report issues.
[...]

PR #378, opened 2019-04-03
Show prompt to enable "fast mode" on large projects

rafeca:
### Description of the Change

This PR adds a prompt to the fuzzy finder which will inform users about the "fast mode" and will allow them to enable it easily.

In order to get the major benefits, the prompt will only be shown when opening large projects, and users will be able to disable the setting via the same prompt later.

Screen Shot 2019-04-03 at 19 54 49

### Alternate Designs

More information in the associated task: https://github.com/atom/fuzzy-finder/issues/376

### Next steps

Things that need to be implemented still:

- [x] Add logic to show the prompt only for large projects.
- [x] Add logic to modify the user settings when clicking the notification.
- [x] Add logic to remove the "new" label after some time (or after the prompt has been shown a specific number of times).
- [x] Add the prompt that allow users to disable the setting once it's enabled.
  - [x] Add disable notification which allows to turn the setting off and show information to provide feedback.
  -  [x] Add logic to modify the user settings when clicking the notification.
- [x] Add metrics.
  - [x] Number of times the "enable prompt" is shown.
  - [x] Number of times people click on the "enable prompt".
  - [x] Number of times people enable the fast mode from the notification.
  - [x] Number of times the "disable prompt" is shown.
  - [x] Number of times people click on the "disable prompt".
  - [x] Number of times people disable the fast mode from the notification.
- [x] Improve codestyle.
- [x] Improve design.
- [x] Test the fuzzy-finder design updates in all UI themes that ship with Atom and also in the [top ~3 community UI themes](https://atom.io/themes/list?sort=downloads&direction=desc).
- [x] Improve prompt and notification copy.

### Verification steps

- [x] Verify prompt with the default Atom config.
  - [x] Check that no prompt is shown on small (< 10k files) projects.
  - [x] Check that the enable prompt is shown on large (> 10k files) projects.
  - [x] Check that the prompt is not shown at the same time than the following messages:
     - [x] Re-indexing project
     - [x] No results found.
  - [x] Check that the prompt automatically appears after the project is reindexed.
  - [x] Check that clicking the prompt shows a notification to enable fast mode.
     - [x] Clicking on the cancel button of the notification does not do anything.
     - [x] Clicking on the confirmation button of the notification changes the ripgrep and the scoring system config values of the fuzzy finder and shows a confirmation message.
- [x] Verify prompt when fast mode is enabled.
  - [x] Check that now the prompt is shown on any project size.
  - [x] Check that the message is different than the one to enable the prompt.
  - [x] Check that the prompt is not shown at the same time than the following messages:
     - [x] Re-indexing project
     - [x] No results found.
  - [x] Check that the prompt automatically appears after the project is reindexed.
  - [x] Check that clicking the prompt shows a notification to disable fast mode.
     - [x] Clicking on the cancel button of the notification does not do anything.
     - [x] Clicking on the confirmation button of the notification changes the ripgrep and the scoring system config values of the fuzzy finder and shows a confirmation message.
- [x] Test the fuzzy-finder design updates in all UI themes that ship with Atom and also in the [top ~3 community UI themes](https://atom.io/themes/list?sort=downloads&direction=desc).
- [x] Verify that each of the newly created metric counters are sent correctly.

### Applicable Issues

https://github.com/atom/fuzzy-finder/issues/376

### DIFF
--- a/lib/experiment-prompt.js
+++ b/lib/experiment-prompt.js
@@ -108,8 +108,8 @@ function renderExperimentPrompt (metricsReporter) {
     return (
       $.span({className: 'experiment-prompt'}, [
         $.a({onmousedown: disableExperimentalFuzzyFinder.bind(null, metricsReporter)}, [
-          $.span({className: ''}, 'Having issues with the fast mode?'),
-          $.span({className: 'icon icon-rocket'}, '')
+          $.span({className: 'icon icon-rocket'}, ''),
+          $.span({className: ''}, 'Having issues with the fast mode?')
         ])
       ])
     )
@@ -121,8 +121,8 @@ function renderExperimentPrompt (metricsReporter) {
     $.span({className: 'experiment-prompt'}, [
       $.a({onmousedown: enableExperimentalFuzzyFinder.bind(null, metricsReporter)}, [
         shouldShowBadge() ? $.span({className: 'badge badge-info'}, 'NEW') : null,
-        $.span({className: ''}, 'Try experimental fast mode?'),
-        $.span({className: 'icon icon-beaker'}, '')
+        $.span({className: 'icon icon-beaker'}, ''),
+        $.span({className: ''}, 'Try experimental fast mode?')
       ])
     ])
   )
--- a/lib/experiment-prompt.js
+++ b/lib/experiment-prompt.js
@@ -122,7 +122,7 @@ function renderExperimentPrompt (metricsReporter) {
       $.a({onmousedown: enableExperimentalFuzzyFinder.bind(null, metricsReporter)}, [
         shouldShowBadge() ? $.span({className: 'badge badge-info'}, 'NEW') : null,
         $.span({className: ''}, 'Try experimental fast mode?'),
-        $.span({className: 'icon icon-microscope'}, '')
+        $.span({className: 'icon icon-beaker'}, '')
       ])
     ])
   )
--- a/lib/experiment-prompt.js
+++ b/lib/experiment-prompt.js
@@ -22,26 +22,25 @@ function shouldShowBadge () {
   return now - firstTimeShown <= TimeToShowNewBadge
 }
 
-function enableExperimentalFuzzyFinder () {
+function showNotification (message, {detail, confirmBtn, cancelBtn, confirmFn, cancelFn}) {
   const notification = atom.notifications.addInfo(
-    'The project you\'ve opened is quite large, which may cause performance issues on the quick open menu. Do you want to enable the new experimental fast mode for the quick open menu?',
+    message,
     {
-      detail: 'This mode can be disabled from the fuzzy finder settings later',
+      detail,
       dismissable: true,
       buttons: [
         {
-          text: 'Enable experimental fast mode',
+          text: confirmBtn,
           onDidClick: () => {
-            atom.config.set('fuzzy-finder.scoringSystem', SCORING_SYSTEMS.FAST)
-            atom.config.set('fuzzy-finder.useRipGrep', true)
-
+            confirmFn && confirmFn()
             notification.dismiss()
           },
           className: 'btn btn-info btn-primary'
         },
         {
-          text: 'No',
+          text: cancelBtn,
           onDidClick: () => {
+            cancelFn && cancelFn()
             notification.dismiss()
           }
         }
@@ -51,7 +50,56 @@ function enableExperimentalFuzzyFinder () {
   )
 }
 
+function enableExperimentalFuzzyFinder () {
+  showNotification(
+    'Do you want to enable the new experimental fast mode for the quick open menu?',
+    {
+      detail: 'This mode can be disabled from the fuzzy finder later',
+      confirmBtn: 'Enable experimental fast mode',
+      confirmFn: () => {
+        atom.config.set('fuzzy-finder.useRipGrep', true)
+        atom.config.set('fuzzy-finder.scoringSystem', SCORING_SYSTEMS.FAST)
+      },
+      cancelBtn: 'No, thanks',
+    }
+  )
+}
+
+function disableExperimentalFuzzyFinder () {
+  showNotification(
+    'Do you want to enable the new experimental fast mode for the quick open menu?',
+    {
+      detail: 'You can reenable it later from the fuzzy finder',
+      confirmBtn: 'Disable experimental fast mode',
+      confirmFn: () => {
+        atom.config.set('fuzzy-finder.useRipGrep', false)
+        atom.config.set('fuzzy-finder.scoringSystem', SCORING_SYSTEMS.ALTERNATE)
+      },
+      cancelBtn: 'No, thanks'
+    }
+  )
+}
+
+
+function isFastModeEnabled () {
+  return (
+    atom.config.get('fuzzy-finder.scoringSystem') === SCORING_SYSTEMS.FAST &&
+    atom.config.get('fuzzy-finder.useRipGrep') === true
+  )
+}
+
 function renderExperimentPrompt () {
+  if (isFastModeEnabled()) {
+    return (
+      $.span({className: 'experiment-prompt'}, [
+        $.a({onmousedown: disableExperimentalFuzzyFinder}, [
+          $.span({className: ''}, 'Having issues with the fast mode?'),
+          $.span({className: 'icon icon-rocket'}, '')
+        ])
+      ])
+    )
+  }
+
   return (
     $.span({className: 'experiment-prompt'}, [
       $.a({onmousedown: enableExperimentalFuzzyFinder}, [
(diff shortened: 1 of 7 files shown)


rafeca:
@jasonrudolph I'm gonna start working now on the first two items on the list:

> - Add logic to show the prompt only for large projects
> - Add logic to remove the "new" label after some time (or after the prompt has been shown a specific number of times).

rafeca:
@jasonrudolph for the "Add logic to show the prompt only for large projects." item I've decided to use the number of files as the threshold instead of the time to crawl: it's way easier to get this number from the view and even if using the time would be a better proxy to know if the user would benefit from the fast mode, the number of files is a good enough proxy IMO.

I've chosen 10K as the threshold, which is a little bit less than what we define as a medium project ([more info](https://github.com/atom/fuzzy-finder/pull/369)). For this project size my maxed-out last-generation macbook pro takes ~4-5s to crawl so the impact of the fast mode will be already quite noticeable.

rafeca:
The main functionality and metrics are already done. This is a screencast of how the general flow looks like:

prompt

The next thing is to improve the messaging and tweak the design if needed, @jasonrudolph do you want to take over the PR for that?

/cc @simurai , @asheren if you want to weight in about the overall experience of the prompt (there's more context about this on https://github.com/atom/fuzzy-finder/issues/376).

rafeca:
My first initial thought is that we should add a confirmation message after enabling/disabling the fast mode... 😅

jasonrudolph:
This is coming together nicely! :zap:

> My first initial thought is that we should add a confirmation message after enabling/disabling the fast mode... 😅

I agree. 👍 

> The next thing is to improve the messaging and tweak the design if needed, @jasonrudolph do you want to take over the PR for that?

I'm happy to do that, but I want to be sure to wrap up my current work-in-progress first. With that in mind, it might be a few days before I can circle back around to this. So, if you want to push it forward in the meantime, feel free. Otherwise, I'll pick this up as soon as I tie up other loose ends. 🙂

rafeca:
> I'm happy to do that, but I want to be sure to wrap up my current work-in-progress first. With that in mind, it might be a few days before I can circle back around to this. So, if you want to push it forward in the meantime, feel free. Otherwise, I'll pick this up as soon as I tie up other loose ends. 🙂

I think that's fine 😄

lib/experiment-prompt.js
```diff
@@ -0,0 +1,133 @@
...
+  )
+}
+
+function disableExperimentalFuzzyFinder (metricsReporter) {
+  metricsReporter.incrementCounter('click-disable-prompt')
+
+  showNotification(
+    'Do you want to disable the new experimental fast mode for the fuzzy finder?',
+    {
+      description: 'If the experimental fast mode is negatively impacting your experience, please [**let us know**](https://github.com/atom/fuzzy-finder/issues/new?title=[Feedback]+Experimental+fast+mode).\n\n(You can reenable fast mode later from the fuzzy finder.)',
```
jasonrudolph: @rafeca: 3bc0884c592e95af179249d314e242a045165533 encourages users to let us know if they're having issues with the experimental fast mode:

Screen Shot 2019-04-09 at 7 34 25 PM

Currently, following that link will open the form for creating a new issue:

https://github.com/atom/fuzzy-finder/issues/new?title=[Feedback]+Experimental+fast+mode

Alternatively, we could start a thread on https://discuss.atom.io/, include a link to that thread, and encourage users to post a comment describing any any problems they encounter. That might help us keep the GitHub issues a bit tidier. What do you think?

jasonrudolph:
> * [ ] Improve design.
> * [ ] Improve prompt and notification copy.

@rafeca: I'm feeling pretty good about the design and the copy at this point. How are things looking to you? You can see an updated walkthrough of the prompts below...

demo

50Wliu:
Is there a reason the prompt is right-aligned?

lib/experiment-prompt.js
rafeca: I think I prefer the idea of using a single place were people can comment, this way we don't have to handle all the issues separately and it'll be easier to group the feedback.

Also, instead of using a discuss thread we can create a "master issue" on GitHub for feedback, so people can comment on the issue the same way they would do on discuss: for me it would be (slightly) more natural to monitor a GitHub issue than a thread on discuss.

Anyways, both options are good 😄

rafeca:
> @rafeca: I'm feeling pretty good about the design and the copy at this point. How are things looking to you? You can see an updated walkthrough of the prompts below...

Oh man, it looks much better! 😍I love the tone used on the messaging!

> Is there a reason the prompt is right-aligned?

Initially we talked about showing it right-aligned to be able to display it at the same time that other messages (like the "re-indexing project..." one), but we finally added some logic to only show the prompt when there's no other message.

From my personal experience using a version of Atom with the left-aligned prompt on the fuzzy finder for a couple of days , when I search stuff my eyes go from the text input to the first result, passing through the prompt text every time causing (a very small) distraction.

By being on the right side, that text is no longer on the regular path of my vision (but it's still very noticeable the first time), which I think it's a better experience.

Of course, I'm not a designer and these are just my personal feelings 😅, so if there's a more technical or funded reason to keep it on the left side I'm ok changing it.

> Was wondering if it should be at the bottom of the list, so it's not the first thing you see every time. But should be fine as an introduction. Maybe later it could be moved.

This would even be a better solution IMHO, unfortunately all the text placeholders available on [`atom-select-list`](https://github.com/atom/atom-select-list) (the component we use for the modal) show the text above the list, and we preferred to not modify that core component.

What do you think about the change that @jasonrudolph has just done? it moves the prompt to the right so it's much less "in your face".

rafeca:
I've added some verification plan to the issue which I'm going to follow. Something I realized while writing it is that since the prompt is only shown on large projects but the fast mode affects any project no matter the size, if somebody enables it while working on a large project but then experiences issues later when working on a smaller project they are not going to be able to disable it easier.

To solve this I'm going to display the opt-out prompt always, no matter the project size (this has the tradeoff of having more "noise" on small projects when the fast mode is enabled, but I think it's worth it).

I'm gonna add a config option to hide the prompt forever to solve this tradeoff.

rafeca:
I've tested the prompt across different themes and there's one of the top community themes that does not display it: [`nucleus-dark-ui`](https://github.com/ignism/nucleus-dark-ui).

Screen Shot 2019-04-10 at 12 36 14

The prompt is not shown but it causes some whitespace to appear between the text input and the list of results.

This is caused by [an overlay div](https://github.com/ignism/nucleus-dark-ui/blob/d6b6b50a0e0bdba50e7c5c6d78ee7e13a0a49e8b/styles/overlays.less#L22-L33 ) that the theme has, which also prevents any other message to appear on the fuzzy finder (like the "reindexing project...").

This is the [4th most downloaded UI theme](https://atom.io/themes/list?sort=downloads&direction=desc), but it's no longer under development based on the Readme and it has [similar issues](https://github.com/ignism/nucleus-dark-ui/issues/114) on other packages.

We can either ignore this issue (it does not have a big impact IMO) or add some logic to disable the prompt if the user has this package.

rafeca:
> We can either ignore this issue (it does not have a big impact IMO) or add some logic to disable the prompt if the user has this theme enabled.

Since it was really easy to do, I've added a check to disable the prompt for that theme

lib/experiment-prompt.js
jasonrudolph: > instead of using a discuss thread we can create a "master issue" on GitHub for feedback

Cool. Let's do that. I'll open a "feedback" issue today and update this link to point to that issue.

jasonrudolph:
I've added some verification plan to the issue which I'm going to follow.

@rafeca: Nice work. Thanks for doing that.

How confident are you that the metrics are being captured correctly? Should that be something we include in the verification plan?

rafeca:
> How confident are you that the metrics are being captured correctly? Should that be something we include in the verification plan?

Ups, yeah I forgot to add this to the verification plan 😅 

Regarding the metrics, we need to do some adjustments in our backend pipelines to handle the counters, I spoke with @telliott27 last week about it and he mentioned that the change would be very easy to do. @telliott27 do you need an issue to be created for that?

lib/experiment-prompt.js
jasonrudolph: > I'll open a "feedback" issue today and update this link to point to that issue.

Done in 653b5b73023a8c8459440962f39caa68a22f0623.

rafeca:
I've verified that the following measures are sent to our backend system after opening/closing the fuzzy finder a few times:

```json
"measures": {
  "show-enable-prompt": 2,
  "click-enable-prompt": 2,
  "cancel-enable-prompt": 1,
  "confirm-enable-prompt": 1,
  "show-disable-prompt": 1,
  "click-disable-prompt": 3,
  "cancel-disable-prompt": 2,
  "confirm-disable-prompt": 1
}
```

Something worth to mention is that the `show-enable-prompt` and `show-disable-prompt` are counted every time the `renderExperimentPrompt()` method is called. Since there's some caching in the rendering logic of the fuzzy finder, when opening two times the fuzzy finder without interacting with it, it's going to count only a single event.

Issue #382, opened 2019-04-30
Fuzzy finder does not handle well results from multiple projects


rafeca:
### Description

When two results from different projects that are located in the same relative path have to be shown, fuzzy finder only shows the result from the first project

### Steps to Reproduce

1. Open two projects containing a `package.json` file in its root directory.
2. Open the fuzzy finder and type `package.json`

**Expected behavior:** Get two results, one from each project.

**Actual behavior:** Get two duplicated results, from the same project.

**Reproduces how often:** Always

PR #383, opened 2019-05-01
Handle correctly similar results from multiple projects

rafeca:
### Requirements

### Description of the Change

This PR fixes https://github.com/atom/fuzzy-finder/issues/382 by adding any root that contains a filter result to the list of final results. 

### Alternate Designs

We could change `fuzzy-native` to accept an object that contains both the full path of the result (or an id) + the string to use for the matcher (as the standard fuzzy matcher does), but that would require sending a much bigger object to the native side which could slowdown thing (plus the change would be bigger and affect other repos).

### Benefits

Fixes https://github.com/atom/fuzzy-finder/issues/382

### Possible Drawbacks

None that I can think of.

### Applicable Issues

https://github.com/atom/fuzzy-finder/issues/382

### DIFF
--- a/lib/fuzzy-finder-view.js
+++ b/lib/fuzzy-finder-view.js
@@ -390,27 +390,32 @@ module.exports = class FuzzyFinderView {
     return {uri: filePath, filePath, label}
   }
 
-  getAbsolutePath (relativePath) {
-    const directories = atom.project.getDirectories()
+  parseResultsFromFuzzyNative (results) {
+    const absolutePaths = new Set()
 
-    if (directories.length === 1) {
-      return path.join(directories[0].getPath(), relativePath)
-    }
+    for (const {value: relativePath} of results) {
+      const directories = atom.project.getDirectories()
 
-    // Remove the first part of the relative path, since it contains the project
-    // directory name if there are many directories opened.
-    relativePath = path.join(...relativePath.split(path.sep).slice(1))
+      if (directories.length === 1) {
+        absolutePaths.add(path.join(directories[0].getPath(), relativePath))
+      } else {
+        // Remove the first part of the relative path, since it contains the project
+        // directory name if there are many directories opened.
+        const newRelativePath = path.join(...relativePath.split(path.sep).slice(1))
 
-    for (const directory of directories) {
-      const absolutePath = path.join(directory.getPath(), relativePath)
+        for (const directory of directories) {
+          const absolutePath = path.join(directory.getPath(), newRelativePath)
 
-      if (fs.existsSync(absolutePath)) {
-        return absolutePath
+          if (fs.existsSync(absolutePath)) {
+            absolutePaths.add(absolutePath)
+          }
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


lib/fuzzy-finder-view.js
```diff
@@ -390,27 +390,32 @@ module.exports = class FuzzyFinderView {
...
+        const newRelativePath = path.join(...relativePath.split(path.sep).slice(1))
 
-    for (const directory of directories) {
-      const absolutePath = path.join(directory.getPath(), relativePath)
+        for (const directory of directories) {
+          const absolutePath = path.join(directory.getPath(), newRelativePath)
 
-      if (fs.existsSync(absolutePath)) {
-        return absolutePath
+          if (fs.existsSync(absolutePath)) {
```
nathansobo: I realize that we this `existsSync` call wasn't introduced with this change, but I really think we should avoid performing synchronous I/O on the main thread. You never know how long the disk read is going to block for, especially on a network drive. One way to do this would be to retain the absolute path of the results returned from `fuzzy-native`. I'm not familiar with its current API, so this may require changes to that module. I doubt they would be too difficult and I am happy to help out with it.

lib/fuzzy-finder-view.js
```diff
@@ -390,27 +390,32 @@ module.exports = class FuzzyFinderView {
     return {uri: filePath, filePath, label}
   }
 
-  getAbsolutePath (relativePath) {
-    const directories = atom.project.getDirectories()
+  parseResultsFromFuzzyNative (results) {
```
nathansobo: This is minor, but I think "parse" isn't really the right word here. Maybe `convertFuzzyNativeResultsToViews` or something along those lines?

spec/fuzzy-finder-spec.js
```diff
@@ -351,6 +360,44 @@ describe('FuzzyFinder', () => {
 
             expect(Array.from(projectView.element.querySelectorAll('li')).filter(a => a.textContent.includes('child-file.txt')).length).toBe(1)
           })
+
+          it('Return all the results if they have the same relative path across multiple root folders', async () => {
```
nathansobo: I generally prefer test descriptions to read like an english sentence when appended to "it":

```
it("returns all the results if..."
```

lib/fuzzy-finder-view.js
rafeca: Good point! I briefly looked at using the async version but that would require changes to `atom-select-list` since it currently expects the `filterFn` function [to be sync](https://github.com/atom/atom-select-list/blob/master/src/select-list-view.js#L293), and changing that was not trivial.

I think that what you mention is similar to the alternate design on the description: that can definitely be done and in general terms it would be a better solution, we just need to make sure that the additional payload that we're serializing in order to pass it to the native module does not regress the overall filtering performance.

nathansobo: Oh, I *totally* missed your comments in the alternate design section. Sorry about that. I think it's worth pursuing that route fairly soon.

Issue #390, opened 2019-05-16
Suggestion: Make the fuzzy finder classes exposed/exported for other plugins to use


robobenklein:
I've been working on some Atom plugins recently and I wanted a fuzzy-finder that I could insert into different places as wanted, however the code here didn't lend itself too openly to that.

Would you accept changes that made it easier for other plugins to gain access to classes such as `ProjectView`, `FuzzyFinderView`, and `BufferView`?

My current idea is that these could either be exported in the package's main module as-is, and potentially additional constructors to aid creating them to avoid the recent issue where the constructors now require the `metricsReporter` to function.

Adding these changes into the upstream would allow metrics reporting to be done in use cases outside the scope of this plugin, which may or may not be desirable.

Two examples I think are clear:

### switch-header-source

https://github.com/dschwen/switch-header-source/issues/36

Using the `require(module.path + /lib/X.js` hack/workaround to gain access to the Fuzzy Finder, would be easier if this plugin provided access or even easier if this plugin provided a factory function for getting new instances of a properly constructed FuzzyFinderView. (with metrics linked, etc)

There's also a lot of other calls to this packages functions which would be massively simplified 

### CodeRibbon

A plugin I'm working on is using a fuzzy-finder as the background element for Panes which don't have any items in them. This allows users to keep the tree view closed more often and still easily see a list of recent / active files in the project.

In my case I extended some classes from this plugin (also using the `require(modulepath + file)` hack) to get rid of the use of `metricsReporter` since I didn't have access to that variable when constructing new `FuzzyFinderView`s.

Extending the FuzzyFinderView: https://github.com/utk-se/CodeRibbon/blob/9ad197cf915a75cd62cdc929e3a564863581896f/lib/fuzzyfinder/fuzzy-finder-view.js

Using a fuzzy finder instance per pane: https://github.com/utk-se/CodeRibbon/blob/9ad197cf915a75cd62cdc929e3a564863581896f/lib/code-ribbon-patch.js#L359-L369

### others

There are probably some other packages I don't know about that also either are currently using classes from this plugin, or would greatly benefit from being able to use a fuzzy finder that's already well known and tested.

I'm willing to make a PR with whatever changes we'd think would best solve this problem - but I'm still unsure of a few things:

 - Should we expose/export the `metricsReporter` instance that fuzzy-finder creates, in order to allow use of the exported class constructors?
[...]

PR #391, opened 2019-05-16
Adds getters for View classes

robobenklein:
### Description of the Change

This adds 5 get functions for access to the 4 view classes and the metricsReporter so that other plugins can have easier access to construct their own instances of the 4 view classes.

### Alternate Designs

Some alternate options were covered in #390 

### Benefits

This doesn't change the existing functionality, and still maintains the lazy-load `requires()` style used in the `createX()` functions.

### Possible Drawbacks

I would think we'd want to expose the metricsReporter here, but if not I can instead add checks in the other view classes to special-case the metricsReporter being undefined.

### Applicable Issues

This is one of the ideas from #390

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -86,11 +86,40 @@ module.exports = {
     return paths
   },
 
+  get FuzzyFinderView () {
+    const FuzzyFinderView = require('./fuzzy-finder-view')
+    console.debug("GET FuzzyFinderView:", this, FuzzyFinderView)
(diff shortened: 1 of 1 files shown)


lee-dohm:
As stated in https://github.com/atom/fuzzy-finder/issues/390#issuecomment-494558400, we believe that services are the proper way to create such a system. Additionally, any pull request would require tests to be included in order to be reviewed and accepted.

Thanks very much for your contribution and your effort!

PR #395, opened 2019-05-29
Fix excluding VCS ignored paths based on the flag in core settings when ripgrep is enabled

nikazooz:
### Description of the Change

This fixes a bug when experimental fast mode that uses ripgrep is enabled. Since excluding VCS ignored paths is the default behavior of ripgrep, option to not ignore them was not being passed when needed.

### Alternate Designs

N/A

### Benefits

Excluding VCS ignored paths from fuzzy search will work a bit more consistently when using ripgrep.

### Possible Drawbacks

N/A

### Applicable Issues

N/A

### DIFF
--- a/lib/load-paths-handler.js
+++ b/lib/load-paths-handler.js
@@ -23,6 +23,7 @@ const emittedPaths = new Set()
 class PathLoader {
   constructor (rootPath, ignoreVcsIgnores, traverseSymlinkDirectories, ignoredNames, useRipGrep) {
     this.rootPath = rootPath
+    this.ignoreVcsIgnores = ignoreVcsIgnores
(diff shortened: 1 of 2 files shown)


50Wliu:
Thanks @nikazooz! Can you please add a test for this to make sure it doesn't regress in the future?

nikazooz:
@50Wliu I don't have much experience with writing tests for atom packages, but this should cover the case this PR solves. I'm open to suggestions for improvement if you have any :slightly_smiling_face:

nikazooz:
Glad I was able to help 🙂
Thank you all for your work on this editor!

50Wliu:
Travis is having rate limit issues, so I think this is fine to merge.

Issue #379, opened 2019-04-10
[Feedback] Experimental Fast Mode


jasonrudolph:
This issue is a placeholder for now. I'll follow up :soon: to fill it out with more details. 😇

visheshvvs:
Tried the experimental mode for my Django project. Rather than having my custom made python file in the search suggestions, it pointed me to Django's internal project files. Somehow it gives them priority over my custom file? Love the concept, but didn't work for me.

rafeca:
Thanks for posting @visheshvvs ! We want to make the fast mode work for everybody, so this feedback is really appreciated.

Can you give me a bit more information about the text that you used for the search, some of the paths that appeared first and the paths that you expected to appear?

Even a couple of screenshots showing the fuzzy finder and its results with the fast mode enabled and disabled will do the trick 😃

**Note**: As a workaround, if you don't want any django file to appear in the fuzzy finder you can ignore the `django` folder from the fuzzy finder package settings:

<img width="821" alt="Screenshot 2019-05-17 at 15 34 45" src="https://user-images.githubusercontent.com/408035/57931747-6927a700-78b9-11e9-9ce7-95a7d8cc3e22.png">

visheshvvs:
@rafeca Thanks for that tip, added **lib** as an excluded directory and I'm no longer seeing those files. 
Just FYI, here's the search results with the experimental mode on without the excluded directory. 
![Screen Shot 2019-05-17 at 10 44 32 AM](https://user-images.githubusercontent.com/5675991/57936148-4d100f80-7891-11e9-959c-ee15611fe2ed.png)

aviatesk:
Thanks for this improvement !

Though with my Atom v1.38.1 x64 on Windows 10, the fast mode of fuzzy finder does NOT seem to ignore VCS-ignored files even though I have the config setting to ignore them.

If I switch back to the previous mode (`alternate`), then they get ignored again (actually it needs reloading Atom, thus it seems that this behaviour is caused at package-loading time).

Is this behaviour intentional ??

rsese:
Hmm, I'm unable to reproduce on Windows 10 with 1.38.1 @aviatesk.  Can you share a minimal example where you see this behavior?

I tested with a test repo with this `.gitignore`:

```
test.txt
```

And 2 files, `test.txt` and `test2.txt`.  With fast mode enabled, when I open fuzzy finder I see `test2.txt` and I don't see `text.txt` as expected.

[...]

PR #399, opened 2019-06-18
Ignore VCS directories even when they are not in ignoreNames

aviatesk:
Hi, this is my very first PR for Atom's core system ! 😁 

### Rationale

If an user's `core.ignoredNames` is modified so that it doesn't include `.git` or `.hg` (they are included by default) but still `core.excludeVcsIgnoredPaths` is `true`, then the new `fast` mode using `ripgrep` will show files in those directories (like `HEAD`, `COMMIT_EDITMSG` and so on).

Because those files are usually regarded as _VCS-ignored_ files and more over, the previous `alternative` mode does exclude those files when `core.excludeVcsIgnoredPaths` is `true`, I believe these files are better to be ignored in `fast` mode as well.

Please refer to https://github.com/atom/fuzzy-finder/issues/379#issuecomment-502635850 for more detail.

### Description of the Change

Only 5 line changes: Add `ripgrep` arguments to exclude those files when `core.excludeVcsIgnoredPaths` is `true`.

### Possible Drawbacks

To include those files, an user need to disable both `core.ignoredNames ` and `core.excludeVcsIgnoredPaths`, and it might be confusing ?

### DIFF
--- a/lib/load-paths-handler.js
+++ b/lib/load-paths-handler.js
@@ -68,6 +68,11 @@ class PathLoader {
         args.push('-g', '!' + ignoredName.pattern)
       }
 
+      if (this.ignoreVcsIgnores) {
+        if (!args.includes('!.git')) args.push('-g', '!.git')
+        if (!args.includes('!.hg')) args.push('-g', '!.hg')
+      }
+
       let output = ''
       const result = childProcess.spawn(realRgPath, args, {cwd: this.rootPath})
 
--- a/spec/fuzzy-finder-spec.js
+++ b/spec/fuzzy-finder-spec.js
@@ -1670,6 +1670,21 @@ describe('FuzzyFinder', () => {
               expect(Array.from(projectView.element.querySelectorAll('li')).find(a => a.textContent.includes('file.txt'))).toBeDefined()
             })
           })
+
+          describe('when core.ignoredNames does not have .git in its glob patterns', () => {
+            beforeEach(() => {
+              atom.config.set('core.ignoredNames', [])
+            })
+
+            it('still ignores .git directory', async () => {
+              await projectView.toggle()
+
+              await waitForPathsToDisplay(projectView)
+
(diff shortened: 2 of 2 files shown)


rafeca:
Wow, this was fast!! Thanks a lot for the PR! 😍

Don't worry about the failure on travis, it's not caused by this PR.

Can you also add a test, so this behaviour does not get broken in the future? You can add it just below this test:

https://github.com/atom/fuzzy-finder/blob/f8e0cb14aa5fa43e6d477f1656b80646a2ff935b/spec/fuzzy-finder-spec.js#L1659-L1672

You can use that test as an inspiration, and do something like `atom.config.set('core.ignoredNames', [])` at the beginning to be sore that the `ignoredNames` param does not have the `.git` folder in it.

> To include those files, an user need to disable both core.ignoredNames and core.excludeVcsIgnoredPaths, and it might be confusing ?

I think it's such a rare case to want to search inside the `.git` folder that this should be fine. Alternatively once this PR lands we can remove the `.git` folder from the `core.ignoredNames` default value.

aviatesk:
I'm welcome to write specs ! 😄 
Not too much sure about writing specs, but there seem a lot of examples in fuzzy-finder, thus I may make it through.

> I think it's such a rare case to want to search inside the .git folder that this should be fine. Alternatively once this PR lands we can remove the .git folder from the core.ignoredNames default value.

Yeah, I totally agree with you and the alternative also sounds sane.

aviatesk:
@rafeca 
I added the spec just now !
Confirmed the current master branch fails the spec I've just added.

I've not created a spec for a case when `core.excludeVcsIgnoredPaths` is set to `false` cause I thought it would make specs more verbose.
But if you want, I'm okay to add a spec for the case 😃

aviatesk:
Thanks again for your help !

I'm really happy to be able to contribute to Atom, my favourite editor, and yeah, will make more in the future ! 😁 

Looking forward to the next release !

PR #406, opened 2019-11-02
Consume STDERR from ripgrep to allow indexing process to complete

fredden:
### Description of the Change
When indexing a project with a lot of broken symbolic links, the indexing process never completes. This is because the `ripgrep` process has written more data to its error output (STDERR) than the calling process has buffer to receive; until this buffer is cleared, the process will not terminate. The solution to this is to process the data on the receiving side.

This pull request adds a simple no-op handler for anything sent back on STDERR from `ripgrep`.

### Alternate Designs
I considered logging the error output from `ripgrep`. This produced a lot of noise when running the test-suite and is unlikely to be of use for anyone in the real world. Logging the error messages returned from `ripgrep` is a trivial change.

### Benefits
Indexing process always completes.

### Possible Drawbacks
Diagnosing some issues may be difficult due to the lack of logging, however before this change any messages sent to STDERR were already being lost.

### Applicable Issues
No obviously related open issues in this repository at time of opening this pull request. Lots of seemly-related external results when searching for this problem on the web.

### DIFF
--- a/lib/load-paths-handler.js
+++ b/lib/load-paths-handler.js
@@ -84,6 +84,9 @@ class PathLoader {
           this.pathLoaded(path.join(this.rootPath, file))
         }
       })
+      result.stderr.on('data', () => {
+        // intentionally ignoring errors for now
+      })
       result.on('close', () => {
         this.flushPaths()
         resolve()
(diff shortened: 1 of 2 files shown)


rsese:
Thanks for the contribution!  CI failure looks unrelated so we'll ask the other maintainers to take a look.

darangi:
Thanks for the contributions 🙇  @fredden

BurntSushi:
Small note: if you are dropping stderr anyway, then you can use `--no-messages` (and possibly `--no-ignore-messages`) to tell ripgrep to be more quiet about file system related errors (and errors from parsing gitignore files).

PR #439, opened 2022-10-23
Simplify Code

mauricioszabo:
Currently, Fuzzy-Finder uses THREE different libraries to make filtering. One of them use Native Bindings, which are not context-aware.

So, we either must decide on using a single one, or migrate to Zadeh (the move I actually want to work later).

I'm drafting this PR so this change does not get lost somewhere :)


AlexWayfer:
> @mauricioszabo closed this 22 minutes ago

Why?

mauricioszabo:
Because Atom will never merge this PR - I actually opened in the wrong repo, it was meant to be opened here: https://github.com/pulsar-edit/fuzzy-finder/pull/1

AlexWayfer:
> Because Atom will never merge this PR - I actually opened in the wrong repo, it was meant to be opened here: [pulsar-edit#1](https://github.com/pulsar-edit/fuzzy-finder/pull/1)

Thank you for answer, completely understandable.

